PR #18, opened 2015-03-09
Implement the new and improved iOS state diagram

shankari:
This includes a distance filter on the location tracking, and integration with
push notifications to detect the end of a trip

phone sensing design 1

Bonus fixes:
- add connect.plist to .gitignore
- restructure the code to separate out the state machine from the actions and
  the location delegate
- Make the progression through the state machine explicit using notifications

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 **/*.xcuserdata
 **/*.xcuserstate
 **/xcuserdata/*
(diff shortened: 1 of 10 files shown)


shankari:
@karunmj, in the spirit of collaboration, and of having somebody look through my code as well, would you like to review my changes before I merge them?

PR #19, opened 2015-03-10
Utility Function

jimmydief:
Initial work on the utility function for our trip recommender system.

### DIFF
--- a/recommender/utility_function.py
+++ b/recommender/utility_function.py
@@ -1,0 +1,43 @@
+# Phase 1: Build a model for User Utility Function (per Vij, Shankari)
+# First, for each trip, we must obtain alternatives through some method
+# (currently Google Maps API), alongside the actual trips which were taken.
+# Once we have these alternatives, we can extract the features from each of the
+# possible trips. Now, we utilize logistic regression with either the simple
+# model given by Vij and Shankari, or utilizing their more complicated method.
+# At the end, we want user-specific weights for their trip utility function.
+
+# Phase 2: Choose optimal future trips
+# Once we have determined a utility function for the user, we can evaluate
+# future trips through this metric. This should look similar to our Phase 1,
+# where we can obtain a set of alternative routes, and then choose an optimal
+# trip based on the user's utility function. However, the key difference is that
+# these trips are no longer static--in Phase 1, we can augment each trip with a
+# set of alternative routes. In Phase 2, this learning is time-dependent (think traffic).
+
+from sklearn import linear_model as lm
+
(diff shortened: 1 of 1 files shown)


jimmydief:
@lechiang

shankari:
Couple of high level changes:

- This needs to be submitted to the e-mission-server repo. e-mission-data-collector is for the phone code that collects sensor data.
- Also, we had discussed the new files would be in `CFC_DataCollector`, possibly in something like `CFC_DataCollector/recommender`. The idea is that everything in `CFC_DataCollector` runs in the background without user intervention, while `CFC_WebApp` is the interactive part.

recommender/utility_function.py
```diff
@@ -0,0 +1,43 @@
...
+# trip based on the user's utility function. However, the key difference is that
+# these trips are no longer static--in Phase 1, we can augment each trip with a
+# set of alternative routes. In Phase 2, this learning is time-dependent (think traffic).
+
+from sklearn import linear_model as lm
+
+# return user-specific weights for a given user based on logistic regression on
+# their past trips and potential alternatives
+def calculateWeights(userID):
+  augmented_trips = getAugmentedTripsByUser(userID, tripFilter)
```
shankari: What are augmented_trips? Where would getAugmentedTripsByUser() be defined and what does it do?

recommender/utility_function.py
```diff
@@ -0,0 +1,43 @@
...
+  features = [extractFeatures(trip) for trip in augmented_trips]
+
+  targets = []
```
shankari: Where will you get the targets from?

recommender/utility_function.py
```diff
@@ -0,0 +1,43 @@
...
+# First, for each trip, we must obtain alternatives through some method
+# (currently Google Maps API), alongside the actual trips which were taken.
+# Once we have these alternatives, we can extract the features from each of the
+# possible trips. Now, we utilize logistic regression with either the simple
+# model given by Vij and Shankari, or utilizing their more complicated method.
+# At the end, we want user-specific weights for their trip utility function.
+
+# Phase 2: Choose optimal future trips
+# Once we have determined a utility function for the user, we can evaluate
+# future trips through this metric. This should look similar to our Phase 1,
```
shankari: Also, how do we get future trips?

jimmydief: Future trips would be chosen by examining a user's most common trips and their alternatives then running our utility function on them.

recommender/utility_function.py
jimmydief: That's the part I'm confused about as well.

PR #21, opened 2015-03-16
Convert the ongoing trip points into a stored trip formatted in the move...

shankari:
...s format

And push the trip to the server when we receive the push notification and
detect the trip as complete


iOS/CFC_Tracker/CFC_Tracker/ViewController.m
```diff
@@ -23,7 +27,19 @@ - (void)viewDidLoad {
     [super viewDidLoad];
     self.transitionTable.dataSource = self;
     self.tdsrmCurrState.adjustsFontSizeToFitWidth = YES;
+    
     // Do any additional setup after loading the view, typically from a nib.
+    self.signInViewController = [[SignInViewController alloc] initWithNibName:nil bundle:nil];
+    self.navigationController = [[UINavigationController alloc] initWithRootViewController:self.signInViewController];
```
shankari: Here's where I initialize the navigation controller with the signInViewController as the root view.

iOS/CFC_Tracker/CFC_Tracker/ViewController.m
```diff
@@ -31,6 +47,30 @@ - (void)didReceiveMemoryWarning {
...
+}
+
+- (void)showSignInView:(id)sender
+{
+    /*
+    if ([self.navigationController.viewControllers containsObject:self.signInViewController]) {
+        // the sign in view is already visible, don't need to push it again
+        NSLog(@"sign in view is already in the navigation chain, skipping the push to the controller...");
+    } else { */
+        [self presentViewController:self.navigationController animated:YES completion:^{
```
shankari: and here's where I present it.

PR #24, opened 2015-04-01
Turn on phone-based activity tracking on iOS

shankari:
This commit turns on phone-based activity tracking on iOS.

However, since the CMMotionActivity class has only read-only properties, it was
fairly challenging to unit test it. I did so by creating a class of my own
(EMActivity, for E-Mission activity) that has very similar data. I also put the
data from the CMMotionActivityManager into the database and segment based on
the values in the database.

This allowed me to write a segmentation based unit test, which works.

When I tested it on a real iphone, however, I did not get any activity data for
the past 4 hours in response to my query. This was in spite of the iOS
documentation that said that the phone continuously collects activity data in
the background, even when not requested by any apps.

This needs to be investigated further, but I had to return my borrowed iPhone6.
Fortunately, test phones are "almost here".

In the meanwhile, if there is no activity information, we fall back on sending
the full set of GPS points for the trip to the server, since we should be able
to segment there using Francois' code.

This will also allow us to work on iPhone4+ instead of iPhone5s+ devices.

### DIFF
--- a/iOS/CFC_Tracker/CFC_Tracker.xcodeproj/project.pbxproj
+++ b/iOS/CFC_Tracker/CFC_Tracker.xcodeproj/project.pbxproj
@@ -12,6 +12,7 @@
 		D062E7BB1ABB1039009373BF /* TripDiaryStateMachineTests.m in Sources */ = {isa = PBXBuildFile; fileRef = D062E7BA1ABB1039009373BF /* TripDiaryStateMachineTests.m */; };
 		D062E7BD1AC4CFAE009373BF /* DataUtilsTests.m in Sources */ = {isa = PBXBuildFile; fileRef = D062E7BC1AC4CFAE009373BF /* DataUtilsTests.m */; };
 		D062E7C01AC4D619009373BF /* TestUtils.m in Sources */ = {isa = PBXBuildFile; fileRef = D062E7BF1AC4D619009373BF /* TestUtils.m */; };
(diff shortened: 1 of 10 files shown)


shankari:
One other bonus fix:

- Changed the initialization of the calendar so that it works on iPhone4 as well. The previous code apparently was iOS8 only, and generated the following error on an iPhone4.

        +[NSCalendar calendarWithIdentifier:]: unrecognized selector sent to class 0x3b85818c

shankari:
@agbison, I think that this is the last of the "moves replacement" work for the moment. Can you take a look?

shankari:
@bellettif, @gaukes, @zacatac, @jeffdh5, this is the change I was talking about today.
As you can see, it didn't really seem to work properly on the physical phone, and I had to give the phone back, so I ended up falling back to sending the unsegmented trip. This means that if we can get the segmentation to work properly on the server side, we should be able to support the iPhone5 also.

Gautham, you should be able to install on your phone once I've merged.
Others, let's plan to meet on Thursday and install the app.

shankari:
Ok, so this is not quite perfect yet. The trips are generated fine on the client and the call to push them to the server is invoked, but the call does not appear to complete, although it doesn't give any errors either. The call is made asynchronously, and the completion handler is never invoked.

I am not sure why this is. Debugging in progress now. I don't think that this glitch should hold up this merge since it is not really related to it, but just wanted to let people know that they shouldn't get too excited.

shankari:
Ok, I know what the push bug is, but will have a draw a small state diagram to figure out how to fix it.
I'm going to merge this at around 5pm today so that I can put it into its own pull request.

PR #25, opened 2015-04-22
Push some pending changes/fixes

shankari:
- Push the trips as part of iOS data collection
- Change the android data collection to use balanced power accuracy, with the corresponding lower power drain


shankari:
Waiting for a review from @agbison for a couple of days before merging.

shankari:
The logos are just placeholders to allow me to use testflight for internal testing. I spent the better part of yesterday wrestling with them, though, because I forgot that you have to regenerate the archive once you have added the logos.

#$#%% apple!

PR #27, opened 2015-05-27
Simplify the data collection phone app by skipping authorization

shankari:
When we are connected to a development server, skip the creation of the JWT and just send the unencrypted email address instead. This makes it much easier to start end to end development.

This contains separate commits for the android and iOS sides.


shankari:
Related pull requests in the main phone app:
https://github.com/e-mission/e-mission-phone/pull/45

and on the server.
https://github.com/e-mission/e-mission-server/pull/131

PR #28, opened 2015-05-27
An update to enable the use of remote servers with self-signed certs

scisley:
Here is an update to the data collector that allows one to use a self signed cert. The cert must be name cert.crt and be put in the res/raw folder.


android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/smap/AddDataAdapter.java
```diff
@@ -117,6 +129,7 @@ public void onPerformSync(Account account, Bundle extras, String authority,
         }
         // First, get a token so that we can make the authorized calls to the server
         String userToken = GoogleAccountManagerAuth.getServerToken(mContext, userName);
+		//userToken = null;
```
shankari: nit: remove comment (I will do this during the manual merge)

PR #29, opened 2015-06-10
Stop launching the app screen after reboot

shankari:
It annoys people to launch the app when the phone reboots. Stop doing that.
Also remove some dead code in the Main Activity.

### DIFF
--- a/android/app/src/androidTest/java/edu/berkeley/eecs/cfc_tracker/test/DataUtilsTest.java
+++ b/android/app/src/androidTest/java/edu/berkeley/eecs/cfc_tracker/test/DataUtilsTest.java
@@ -104,7 +104,7 @@ public class DataUtilsTest extends AndroidTestCase {
 	
 	public void testGetJSONPlace() throws JSONException {
 		Location loc = makeLoc(37, -122, 1);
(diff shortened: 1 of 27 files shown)


shankari:
@scisley, as I really focus on solidifying the data collection on the phone, I found it annoying that if things went wrong in the background, but I wouldn't be able to figure out what it was because it would have aged out of the logcat cache by the time I got home and hooked my phone up to adb.

So I set up a filehandler for a custom logger, and logged to it, and then display the logs on the main screen. I'd actually attempted to do this quite a while before, but I wasn't displaying the data or sending it to the server, and it turned out that the file logging wasn't actually working.

So I spent today debugging it and getting it to work properly and making sure that it will continue to work properly by writing a unit test.

Take a look and tell me what you think. While there are existing crash reporting frameworks for android, I am not aware of a logging solution other than http://api.shephertz.com/apis/logging.php. And that solution sucks for us because it logs instantaneously, which will kill your battery faster than the GPS :)

screen shot 2015-06-20 at 11 00 05 pm

shankari:
@scisley, I am going to merge this tonight so that I can move to a branch as I should have done from the very beginning.

PR #30, opened 2015-06-11
Loc fix

scisley:
My data-collector kept dying on my phone (several times a day). I hunted down the error to LocationChangeIntentService where it grabs an extra from the passed in intent but doesn't check that the extra actually exists, so later on you get a NullPointerException when you try to access it. Right now this code just exits if the passed in intent doesn't have the right data. See the SO link in the code for details. (please ignore all the weird commits prior to the final one, it was me fighting git).


shankari:
Great catch! I hate undocumented APIs.. :(
Added issue to investigate this further and determine if/how to use it.
https://github.com/e-mission/e-mission-data-collection/issues/31

For now, it is critical to stop crashing...

PR #32, opened 2015-06-29
Add a design document for the data formats

shankari:


shankari:
A preview of the document is at:
https://github.com/shankari/e-mission-data-collection/blob/switch_to_our_format/doc/README.dataformat.md

shankari:
I've also added the same material to the wiki 
https://github.com/e-mission/e-mission-data-collection/wiki/Data-format-design

Let us see which format is easiest to work with and then stick with it going forward.

shankari:
@scisley, @joshzarrabi, @farietta, @nephraim - here's my current plan for the user cache format. I plan to start making the corresponding changes to the data collection code tomorrow, but they won't be irreversible changes.

Let's discuss the format in the context of this pull request and then in person on Thursday. Steve, let me know if you want to dial in to the Thursday meeting.

scisley:
Thanks Shankari. I'll take a look at this. I'll try to dial in on Thursday.

shankari:
@scisley, @joshzarrabi, @farietta, @nephraim, I am updating the wiki instead of this README because I think that the design decision is too complicated for a simple README. I have made the README simpler and added a link to the wiki instead.

It doesn't look like people are adding comments to the pull request anyway...

PR #33, opened 2015-07-01
Add a new unit test which tests the serialization/deserialization of …

shankari:
…classes using gson


shankari:
I wanted to commit this in the switch_to_our_format branch, but I already have a pending pull request there for the documentation

shankari:
I think that this will be the switch_to_our_format branch until the documentation issue is resolved.

PR #37, opened 2015-07-06
Initial implementation of the user cache

shankari:
Heavily influenced by the ClientStatsHelper, since it essentially does the same thing.
Left to do:
- play around with gson and then GSON-ify the interfaces
- start calling the code from the intents
- change the data sync to use this instead


shankari:
Everything compiles.
Merging now so that I can have a separate pull request just for the poll sensors.

PR #50, opened 2015-09-06
Master

patrickSUP:
Merge of OBD communication with E-Mission

Read the file instructions.txt to get how to use
and which adapters are supported

I inserted in the project the entire library of pires obd-java-api because I needed to correct a few bugs.

Note that it should be used a service instead of the async task to commmunicate with the vehicle, with a handler to retrieve the data when there is a location update.
Now the user cache is updated sending needed data to an intent service, like the location update intent.

The battery drain on fine mode is accettable, during trips of 2 hours the percentage of battery drops of abot 5%

The app need more testing to ensure compatibility with the largest number possible of vehicles.

### DIFF
--- a/android/app/src/main/java/com/github/pires/obd/commands/ObdCommand.java
+++ b/android/app/src/main/java/com/github/pires/obd/commands/ObdCommand.java
@@ -1,0 +1,307 @@
+package com.github.pires.obd.commands;
+
+import com.github.pires.obd.exceptions.*;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.ArrayList;
+
+/**
+ * Base OBD command.
+ */
+public abstract class ObdCommand {
+
+    /**
+     * Error classes to be tested in order
+     */
+    private final Class[] ERROR_CLASSES = {
+            UnableToConnectException.class,
+            BusInitException.class,
+            MisunderstoodCommandException.class,
+            NoDataException.class,
+            StoppedException.class,
+            UnknownErrorException.class,
+            UnsupportedCommandException.class
+    };
+    protected ArrayList<Integer> buffer = null;
+    protected String cmd = null;
+    protected boolean useImperialUnits = false;
+    protected String rawData = null;
+    protected long responseTimeDelay = 200;
+    private long start;
+    private long end;
+
+    /**
+     * Default ctor to use
+     *
+     * @param command the command to send
+     */
+    public ObdCommand(String command) {
+        this.cmd = command;
+        this.buffer = new ArrayList<Integer>();
+        /*if (this instanceof ReturnAsapCommand) {
+            this.cmd += " 1";//speed up
+        }*/
+    }
+
+    /**
+     * Prevent empty instantiation
+     */
+    private ObdCommand() {
+    }
+
+    /**
+     * Copy ctor.
+     *
+     * @param other the ObdCommand to copy.
+     */
+    public ObdCommand(ObdCommand other) {
+        this(other.cmd);
+    }
+
+    /**
+     * Sends the OBD-II request and deals with the response.
+     * <p>
+     * This method CAN be overriden in fake commands.
+     *
+     * @param in  a {@link java.io.InputStream} object.
+     * @param out a {@link java.io.OutputStream} object.
+     * @throws java.io.IOException            if any.
+     * @throws java.lang.InterruptedException if any.
+     */
+    public void run(InputStream in, OutputStream out) throws IOException,
+            InterruptedException {
+        start = System.currentTimeMillis();
+        sendCommand(out);
+        readResult(in);
+        end = System.currentTimeMillis();
+    }
+
+    /**
+     * Sends the OBD-II request.
+     * <p>
+     * This method may be overriden in subclasses, such as ObMultiCommand or
+     * TroubleCodesCommand.
+     *
+     * @param out The output stream.
+     * @throws java.io.IOException            if any.
+     * @throws java.lang.InterruptedException if any.
+     */
+    protected void sendCommand(OutputStream out) throws IOException,
+            InterruptedException {
+        // write to OutputStream (i.e.: a BluetoothSocket) with an added
+        // Carriage return
+        out.write((cmd + "\r").getBytes());
+        out.flush();
+
+    /*
+     * HACK GOLDEN HAMMER ahead!!
+     *
+     * Due to the time that some systems may take to respond, let's give it
+     * 200ms.
+     */
+        Thread.sleep(responseTimeDelay);
+    }
+
+    /**
+     * Resends this command.
+     *
+     * @param out a {@link java.io.OutputStream} object.
+     * @throws java.io.IOException            if any.
+     * @throws java.lang.InterruptedException if any.
+     */
+    protected void resendCommand(OutputStream out) throws IOException,
+            InterruptedException {
+        out.write("\r".getBytes());
+        out.flush();
+    }
+
+    /**
+     * Reads the OBD-II response.
+     * <p>
+     * This method may be overriden in subclasses, such as ObdMultiCommand.
+     *
+     * @param in a {@link java.io.InputStream} object.
+     * @throws java.io.IOException if any.
+     */
+    protected void readResult(InputStream in) throws IOException {
+        readRawData(in);
+        checkForErrors();
+        fillBuffer();
+        performCalculations();
+    }
+
+    /**
+     * This method exists so that for each command, there must be a method that is
+     * called only once to perform calculations.
+     */
+    protected abstract void performCalculations();
+
+    /**
+     *
+     */
+    protected void fillBuffer() {
+        rawData = rawData.replaceAll("\\s", ""); //removes all [ \t\n\x0B\f\r]
+        rawData = rawData.replaceAll("(BUS INIT)|(BUSINIT)|(\\.)", "");
+
+        if (!rawData.matches("([0-9A-F])+")) {
+            throw new NonNumericResponseException(rawData);
+        }
+
+        // read string each two chars
+        buffer.clear();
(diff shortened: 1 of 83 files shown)


android/app/src/main/java/com/github/pires/obd/commands/ObdCommand.java
```diff
@@ -0,0 +1,307 @@
...
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
+import java.util.ArrayList;
+
+/**
+ * Base OBD command.
+ */
+public abstract class ObdCommand {
```
shankari: Are you planning to submit your fixes as patches to the `obd-java-api` project? I think that it would be good to do that so that we can get their ongoing fixes without issues. Otherwise, as we diverge from their codebase, we may run into merge conflicts going forward. It is also generally a Good Thing to do and will look good on your resume :)

android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/CarActivity.java
```diff
@@ -0,0 +1,878 @@
...
+            return true;
+        }
+
+        return super.onOptionsItemSelected(item);
+    }
+
+    /**
+     * Async task responsible for getting updates from adapter and refreshing the user interface
+     */
+    private class GetUpdates extends AsyncTask<Void, String, Void> {
```
shankari: I thought that AsyncTasks were in the scope of their parent activity (context). In particular, I think that this model is broken in the following case. Let's say that the user doesn't want to have a real-time display distracting her while driving, but instead, wants to look at the overview at the end. So she connects her phone but doesn't launch the activity. Then the trip won't be recorded at all because nothing will happen while the activity is suspended or killed. Have you tested this case?

android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/CarActivity.java
```diff
@@ -0,0 +1,878 @@
...
+                    long deltaTime = currentTime - previousTime;
+                    previousTime = currentTime;
+                    engineRpmCommand.run(bluetoothSocket.getInputStream(), bluetoothSocket.getOutputStream());
+                    speedCommand.run(bluetoothSocket.getInputStream(), bluetoothSocket.getOutputStream());
+
+                    rpmResult = engineRpmCommand.getFormattedResult();
+                    speedResult = speedCommand.getFormattedResult();
+                    Log.d(TAG, "deltaTime: " + deltaTime);
+                    //calculation of distance driven
+                    if (fineMode) {
```
shankari: This is pretty ugly. Refactor into an abstract superclass and two concrete subclasses for fine and coarse?

android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/CarActivity.java
```diff
@@ -0,0 +1,878 @@
...
+                            Log.d(TAG, "intermediate tankLevel: " + finalTankLevel);
+                        } else {
+                            Log.d(TAG, "intermediate tankLevel: " + finalTankLevel);
+                        }
+                    }
+
+
+                    //Saving data to external memory for analysis
+                    String currentDateandTime = new SimpleDateFormat("yyyy-MM-dd").format(new Date());
+                    fileUtilities.write("DataEMission" + currentDateandTime + ".txt", engineRpmCommand.getRPM() + "\t" + speedCommand.getMetricSpeed() + "\t" +/*fuelEconomy.getLitersPer100Km()*/fuelEconomy.getFlow() + "\t" + currentTime + "\n");
```
shankari: Why do you need to write to a file? Are you using this elsewhere?

android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/CarActivity.java
```diff
@@ -0,0 +1,878 @@
...
+                    //saving to userCache, consider to use always coarseMode (finemode=false) to limit database updates
+                    //you can also add if trip is ended (low rpm)
+                    Bundle b = toBundle(fuelEconomy.getFlow(),speedCommand.getMetricSpeed(),getDistanceDriven(),engineRpmCommand.getRPM(),getFuelUsed());
```
shankari: I am a bit confused about your structure for the OBD -> user cache interaction. If you recall, my vision of the architecture was that the bluetooth data would be pulled from the location call (specifically here):
https://github.com/e-mission/e-mission-data-collection/blob/master/android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/location/LocationChangeIntentService.java#L60

I remember that we talked about this and that was the motivation for the coarse mode - in particular, that it is consistent with the iOS polling mechanism. Is there a reason why you have not implemented this? Since you are the first user, it would be interesting to get feedback on why you felt that the design was inadequate.

shankari:
I have three main high-level questions that I have replied to inline. Once we resolve these, I can do a more detailed review.

android/app/src/main/java/com/github/pires/obd/commands/ObdCommand.java
patrickSUP: The fact is that I think the developer added the " 1" carachter at the end of every command thinking that this (the speed up) would work with every car/adapter but with mine it doesn't work, so I'm going to ask if my supposition is right

android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/CarActivity.java
patrickSUP: Yes, in my case the async task still works if I exit from the activity, but I think is much more safe to implement something like a service that will stay in background sending commands to the adapter. The real time info was intended for debugging, in the final app it's best not to display such info.

android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/CarActivity.java
patrickSUP: I used it to do offline analisys since when it was a standalone app I recorded the data on text file, now it is not necessary anymore

android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/CarActivity.java
patrickSUP: Yes, that design is fine, I was misled by the name sensor and because I did this part last. What do you think if in the method getAndSaveAllValues() I call a static method that retrieves the values saved so far from the class responsible for handling data? because thats not a real sensor that needs to be polled,I mean it is, but only after having done all the calculations.

android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/CarActivity.java
patrickSUP: Yes, that can be done. Much more elegant

android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/CarActivity.java
patrickSUP: Or maybe you mean something like a software sensor?

android/app/src/main/java/com/github/pires/obd/commands/ObdCommand.java
patrickSUP: I opened an issue:
https://github.com/pires/obd-java-api/issues/67

android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/CarActivity.java
shankari: Yes, this is a software sensor that is calculated from hardware sensor values. The thought is that all the code in the AsyncTask, including reading the raw data and then calculating the values, could happen inside getAndSaveAllValues(). And then, instead of generating a new intent, you could just save the values to the database.

Since this is invoked from an existing IntentService, you should be fine wrt async operation.

See 
https://github.com/e-mission/e-mission-data-collection/blob/master/android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/sensors/BatteryPollSensor.java
for an example.

android/app/src/main/java/com/github/pires/obd/commands/ObdCommand.java
shankari: This has been resolved. Is it OK to now remove the checked-in library code and replace it with a gradle dependency?

android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/CarActivity.java
shankari: Yes, this is a software sensor that is calculated from hardware sensor values. The thought is that all the code in the AsyncTask, including reading the raw data and then calculating the values, could happen inside getAndSaveAllValues(). And then, instead of generating a new intent, you could just save the values to the database.

Since this is invoked from an existing IntentService, you should be fine wrt async operation.

See 
https://github.com/e-mission/e-mission-data-collection/blob/master/android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/sensors/BatteryPollSensor.java
for an example.

android/app/src/main/java/com/github/pires/obd/commands/ObdCommand.java
patrickSUP: Yes, I'll do that

android/app/src/main/java/edu/berkeley/eecs/cfc_tracker/CarActivity.java
shankari: Yes, you really need a service. That's why I have the data polled from the location IntentService.

shankari:
Hi Patrick,

Any updates on the recommended changes?

Thanks,
Shankari

Issue #36, opened 2015-07-06
SharedPreferences are sometimes not saved properly


shankari:
It looks like SharedPreferences do not save properly in some cases.
Here's what I saw from the app logs for a trip in which I left Tied House at 8:30pm and came home.
The app appeared to have been killed and relaunched by the geofence exit, since the id was 9.

    [9|1435807867723|7/1/15 8:31 PM|FINE]TripDiaryStateMachineReceiver : after reading from the prefs, the current state is local.state.waiting_for_trip_start

After processing the geofence exit, the new state was ongoing_trip.

    [14|1435807867853|7/1/15 8:31 PM|FINE]TripDiaryStateMachineReceiver : newState after handling action is local.state.ongoing_trip

However, when the state is next read, it is still waiting_for_trip_start

    [87|1435808413146|7/1/15 8:40 PM|FINE]TripDiaryStateMachineReceiver : after reading from the prefs, the current state is local.state.waiting_for_trip_start

So when the trip end is detected, we think that the current state is already waiting_for_trip_start,
so we do not turn off ongoing tracking, and we do not establish the geofence.

shankari:
Additional logging added in pull request #35

shankari:
A potential solution might be to use commit() instead of apply().
commit() saves synchronously and returns a result value
apply() saves asynchronously and fails silently. I am not sure what might cause the save to fail.
It is not supposed to be the android lifecycle because:
> You don't need to worry about Android component lifecycles and their interaction with apply() writing to disk. The framework makes sure in-flight disk writes from apply() complete before switching states. 

On the other hand, this is a service, not an activity. Maybe that's why it is confusing?

sfwatergit:
If this refers specifically to the call on SharedPrefs in TripDiaryStateMachineReceiver, then it is due to the lifecycle of BroadcastReceiver, which is not a service. You may wish to extend WakefulBroadcastReceiver, use go Async, or simply refactor the BroadcastReceiver as an inner static class of a service, which is a bit more idiomatic in my experience.

shankari:
> If this refers specifically to the call on SharedPrefs in TripDiaryStateMachineReceiver, then it is due to the lifecycle of BroadcastReceiver, which is not a service, but ceases to exist as a separate component almost immediately after completion of any running code.

Yes, it refers to TripDiaryStateMachineReceiver. However, as I quoted above, the documentation for apply() specifically says that
[...]

PR #51, opened 2015-09-12
Switch to handling most of the functionality of the trip diary state …

shankari:
…machine in a separate service

It looks like once the broadcast receiver is marked as inactive (after
returning from onReceive) even callbacks can be terminated in the middle.

This launches a service to do the bulk of the work and avoid early termination.

Testing done: LocationTests run successfully. They don't actually return a pass state because the test conditions need to be modified to match the new trip end detection, but the logs indicate that the services are all started correctly.


shankari:
This resolves issue #36

PR #54, opened 2015-09-29
Change the TripDiary manipulation from an intent service to a service

shankari:
While an IntentService is recommended for performing operations in a separate
background thread, it is not a suitable option for us for the following
reasons:
- we are already running in the background since we are invoked from the BroadcastReceiver
- the IntentService calls stopSelf automatically once the intent is handled.
  But this means that async operations, such as the ones to create the geofence
  or to start tracking may return only after the service has already been
  stopped or started.
    - So in order to get the state machine to work properly, we want to wait
      until all operations have finished, and the new state is set to stop the service


shankari:
jenkins, test this please

PR #130, opened 2016-09-29
Fix corner case data collection issues + iOS8 support

shankari:
More details in individual commit messages

### DIFF
--- a/src/android/ConfigManager.java
+++ b/src/android/ConfigManager.java
@@ -2,27 +2,41 @@ package edu.berkeley.eecs.emission.cordova.tracker;
 
 import android.content.Context;
 
+import com.google.gson.JsonParseException;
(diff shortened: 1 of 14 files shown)


shankari:
@flongo82, https://github.com/e-mission/e-mission-data-collection/pull/130/commits/2da64dd49e4282156b3feaa1f691ac92f3258443 is the set of changes to handle location being turned off, if you want to take a look.

shankari:
@flongo82, I tested this by doing the following:

### Test1 ###
- get the phone into `WAITING_FOR_TRIP_START` state.
- Turn off location services.
- Should get a notification asking you to turn it back on.

### Test 2 ###
- Have network services turned off (FSM will be in start state)
- Try to send an initialize transition
- Get a notification about an error in creating the geofence

You could see if it works too, or make better error messages or ???

PR #146, opened 2016-12-16
Add a timestamp to Transition and Battery objects

shankari:
Before this, we have not used these objects on the client, only on the server.
On the server, we do have a `ts` that we have copied the metadata `write_ts` to.
Now that we want to use the transition objects on the client too, we can either:
a) read the full object, including metadata, which means we can't use the wrappers
b) add a timestamp to the object

b) seems nicer because:
- it allows us to potentially get rid of the formatters later
- allows us to generate and save at different times
- makes all objects more standardized

So that is what we will do!


shankari:
This is used as part of commit https://github.com/e-mission/e-mission-transition-notify/pull/2/commits/a427677288b8865ecb0cdcb1c820f125f12eb4a6, pull request https://github.com/e-mission/e-mission-transition-notify/pull/2

Issue #164, opened 2018-01-30
Trip end never detected


shankari:
At least two users have reported that the trip end is never detected.
And at least one of them reported this even after the trip was manually reset.
This would normally not be a problem, but we don't upload data until the trip has ended.

Let's investigate further before we send this out to hundreds of people.

shankari:
Triangulating on one of the phones, everything was working fine until Dec 22nd, when it abruptly stopped working. Looking through the recent logs, we received activity updates but not location updates for minutes at a time. The problem is with the frequency of location updates.

shankari:
Another thing that I noticed - the phone that I tested was on android 8.0. We have not tested on android 8.0, which apparently includes changes to background location updates.
https://developer.android.com/about/versions/oreo/android-8.0-changes.html

> In order to preserve battery, user experience, and system health, background apps receive location updates less frequently when used on a device running Android 8.0. This behavior change affects all apps that receive location updates, including Google Play services. 

https://developer.android.com/about/versions/oreo/background-location-limits.html

> In an effort to reduce power consumption, Android 8.0 (API level 26) limits how frequently background apps can retrieve the user's current location. Apps can receive location updates only a few times each hour.

Workarounds include:

- Bring your app to the foreground.
   - Non starter
- Start a foreground service in your app by calling startForegroundService(). When such a foreground service is active, it appears as an ongoing notification in the notification area.
   - Might be annoying to see the icon, but might be the easiest short-term option
- Use elements of the Geofencing API, such as the GeofencingApi interface, which are optimized for minimizing power use.
   - Already done
- Use a passive location listener, which may receive faster location updates if there are foreground apps requesting location updates at a faster rate.
   - What if they are not?
- Note: If your app needs access to location history that contains time-frequent updates, use the batched version of the Fused Location Provider API elements, such as the FusedLocationProviderApi interface. When your app is running in the background, this API receives the user's location more frequently than the non-batched API. Keep in mind, however, that your app still receives updates in batches only a few times each hour.
   - this looks like it would be a good long-term solution, but will require some additional testing to see what we actually get in the batch, how often we get it, how to modify the trip end detection appropriately, etc.

shankari:
ok, so next steps are:
- check with the other user that they are also on android 8
- start a new foreground service when we start location tracking

Hopefully, we can finish this by tomorrow night so that Marta's class + study will use correct app.

shankari:
This seems to have fixed the issue on android oreo.
[...]

PR #165, opened 2018-02-20
Turn on a foreground service with associated notification on android 8+

shankari:
Without this fix, we get location updates only once or twice an hour and never
detect that the trip ended.
This is a short-term, potentially permanent, fix for #164

Also bump up the version number to reflect this change.


shankari:
This depends on https://github.com/e-mission/cordova-unified-logger/pull/24

shankari:
Hard to synchronize across repositories, so let's just keep inconsistent window as short as possible.

PR #169, opened 2019-02-18
Fix location services tracking toggle on android

shankari:
This fixes the issue with turning location services off in the `WAITING_FOR_TRIP_START` state.
That generates a tracking error, which then generates a notification asking people to turn on location services again.
Assuming the checkForLocationSettings fails, the generated notification prompts the user to turn on location services again,
and the data collection plugin then re-initializes the FSM so that we return to `WAITING_FOR_TRIP_START`


src/android/location/TripDiaryStateMachineService.java
```diff
@@ -103,6 +112,15 @@ public int onStartCommand(Intent intent,  int flags, int startId) {
             Log.d(this, TAG, "service restarted! need to check idempotency!");
         }
         mTransition = intent.getAction();
+        if (currActions.contains(mTransition)) {
```
shankari: after making sure that we set the new state every time, I am not sure this is actually required. But I am not going to remove this right now due to the tight schedule

shankari:
Has been tested by @PatGendre, @deepalics0044 and @hmharvey, so I assume this is GTG!

PR #177, opened 2020-03-30
Force the gson library to the last valid version that supports Java 1.6

shankari:
This fixes
https://github.com/e-mission/e-mission-docs/issues/475#issuecomment-559623057

@st-patrick, in case you want to review this

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -112,7 +112,7 @@
 		</service>
     </config-file>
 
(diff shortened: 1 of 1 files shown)


shankari:
Do we need to force this, or should we upgrade to a later version of java?
Let's force this for now and upgrade after the current build is complete.

PR #192, opened 2021-06-05
Handle android 11 "always" permission properly + Humanize android notification messages

shankari:
- Make messages shorter
- Make them more human-understandable
- Popup app settings on android 11
  - to avoid issue with having to open settings anyway
  - and to deal with the fact that the app dialog will not pop up if the user select anything other than always the first time

This fixes https://github.com/e-mission/e-mission-docs/issues/608


shankari:
Re-testing done on:
- android 11
- android 10
- iOS 11

PR #195, opened 2021-10-26
Native upgrade 2021

shankari:
- Basic upgrades to all libraries
- Significant refactor of the permission and settings code to make it more step-by-step and to clarify what we want people to do. This is particularly important as we have more and more required permissions.


shankari:
Note that https://github.com/e-mission/e-mission-data-collection/pull/195/commits/61a7b8b99ffecd22acc41e9644429b75f5f8312f also includes the checks to display background restrictions only on android.

PR #208, opened 2023-09-25
Battery Optimization Plugin

louisg1337:
Slightly refactored pre existing code to create a pop up that will prompt users to ignore battery optimizations for the app instead of getting brought to the settings and having to do it themselves. 

There is a slight bug that needs to be fixed before it can be merged. Whenever you prompt this new popup, all the other buttons that change different permissions just don't work anymore. They normally send you to the settings to turn off a specific setting, but now it doesn't send you to the settings, it just lags for a second.

### DIFF
--- a/src/android/verification/SensorControlForegroundDelegate.java
+++ b/src/android/verification/SensorControlForegroundDelegate.java
@@ -428,10 +428,15 @@ public class SensorControlForegroundDelegate {
       Log.i(cordova.getActivity(), TAG, "Battery optimizations enforced, asking user to ignore");
       this.cordovaCallback = cordovaCallback;
       cordova.setActivityResultCallback(plugin);
-      Intent intent = new Intent(Settings.ACTION_IGNORE_BATTERY_OPTIMIZATION_SETTINGS);
-      cordova.getActivity().startActivityForResult(intent, SensorControlConstants.OPEN_BATTERY_OPTIMIZATION_PAGE);
+      Intent intent = new Intent(Settings.ACTION_REQUEST_IGNORE_BATTERY_OPTIMIZATIONS);
+      String packageName = cordova.getActivity().getPackageName();
+      intent.setData(Uri.parse("package:" + packageName));
+      cordova.getActivity().startActivity(intent);
+      cordovaCallback.success("Battery optimization request sent!");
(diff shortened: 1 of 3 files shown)


src/android/verification/SensorControlForegroundDelegate.java
```diff
@@ -418,22 +461,44 @@ public void checkIgnoreBatteryOptimizations(CallbackContext cordovaCallback) {
...
-      Intent intent = new Intent(Settings.ACTION_IGNORE_BATTERY_OPTIMIZATION_SETTINGS);
-      cordova.getActivity().startActivityForResult(intent, SensorControlConstants.OPEN_BATTERY_OPTIMIZATION_PAGE);
+      this.permissionChecker = getPermissionChecker(
+          SensorControlConstants.IGNORE_BATTERY_OPTIMIZATIONS,
+          SensorControlConstants.REQUEST_BATTERY_PERMISSION,
+          "Please turn this on!",
+          "Please reconsider!"
+      );
+      this.permissionChecker.requestPermission();
+      return;
```
shankari, line 480: I think that the issue is that you do have to still open an intent, although the intent shows up as a popup. This is not a permission that can be requested using `requestPermission` (I wasn't super clear about this because I thought that popups were all permissions).

There is an example of how to open an intent from the permission checker here:
https://github.com/e-mission/e-mission-data-collection/blob/c21f8354f133498dedec37aeaf98ec3afddaefec/src/android/verification/SensorControlForegroundDelegate.java#L94

that is where we open the app settings page

you would need to add a new flag for "open optimization checker" or something like that which would open the intent

shankari: I do wonder a bit whether opening intents from the permission checker is worth it, but let's follow the pattern for now and move both intent launches out if it is not worth it for consistency.

shankari (changes requested):
This has a lot of extraneous whitespace so it is very hard to review and will mess up the git blame. I would suggest copying out `SensorForegroundDelegate` to temp or something like that, and then reverting the changes and reintroducing only the changes that are actually relevant.

Once this is cleaned up, it should be fine to merge - I was able to figure out the changes and they look fine. In fact, in a future cleanup, we may want to handle `openAppSettings` in the same way (with checking the permission constant) instead of a separate boolean. I do think that is a lot cleaner.

PR #210, opened 2023-10-20
Location permissions rework

louisg1337:
Created a new location permissions flow to make it easier for users. This change involved sending the user to a different location settings screen, creating helpful popups that explain to the user what needs to be done, and incorporated more meaningful error messages.

### DIFF
--- a/src/android/verification/SensorControlForegroundDelegate.java
+++ b/src/android/verification/SensorControlForegroundDelegate.java
@@ -31,6 +31,8 @@ import android.net.Uri;
 import android.os.AsyncTask;
 import android.os.Build;
 import android.provider.Settings;
+import android.content.DialogInterface;
+import android.app.AlertDialog;
 
 
 import androidx.core.app.ActivityCompat;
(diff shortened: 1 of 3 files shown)


src/android/verification/SensorControlForegroundDelegate.java
```diff
@@ -258,6 +269,76 @@ private void openAppSettingsPage(CallbackContext callbackContext, int requestCod
...
+     * through the old flow of just opening up the app settings page.
+     * 
+     * @param callbackContext
+     * @param requestCode
+     */
+    public void beforeAppSettingsDialog(CallbackContext callbackContext, int requestCode){
+      new AlertDialog.Builder(cordova.getActivity())
+      .setTitle(R.string.location_permission_intermediary_title)
+      .setMessage(R.string.location_permission_settings_message)
+      .setPositiveButton("OK", new DialogInterface.OnClickListener() {
```
shankari, line 284: Whoo! It's been a while since I have worked with the `AlertDialog`

src/android/verification/SensorControlForegroundDelegate.java
```diff
@@ -580,7 +699,10 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
           // this is the activity result callback, so only launched when the app settings are used
           // so we don't need to use the permission checker, we know that the only option
           // is to launch the settings
-          cordovaCallback.error(cordova.getActivity().getString(R.string.location_permission_off_app_open));
+          if (cordovaCallback != null) {
+            String errorMessage = generateLocationPermissionError();
+            cordovaCallback.error(errorMessage);
+          }
```
shankari, line 705: nit: in general, it is not a good idea from a systems perspective to silently ignore an error. You obviously don't want to crash, but you should at least log what happened.

src/android/verification/SensorControlForegroundDelegate.java
```diff
@@ -528,6 +622,31 @@ public void onRequestPermissionResult(int requestCode, String[] permissions,
...
                 this.permissionChecker = null;
                 break;
+            case SensorControlConstants.LOCATION_INTERMEDIARY:
+                // This case allows us to do the 3 step user location permissions setup. ACCESS_FINE_LOCATION gets called first,  
+                // and this is the callback case for it. If it is allowed, prompt user with a dialog, then send them to the location
+                // permissions settings if they accept. 
+                if (grantResults.length == 0) {
+                  // Covers weird error where SensorControlConstants.ENABLE_BACKGROUND_LOC_PERMISSION will get called again after an
+                  // initial intended call. The second erroneous call creates an error because grantResults = [], causing an array 
+                  // out of bounds error, so just return.
```
shankari, line 632: Do you know why this happens? Is it triggered from the UI or from the onboarding in the native code? I am fine with merging this now, but we should file a issue to investigate and clean up.

shankari (approved):
I am merging now, but please file an issue to understand where the second call comes from.

PR #209, opened 2023-09-28
New battery optimization request permissions popup.

louisg1337:
Integrated the `REQUEST_IGNORE_BATTERY_OPTIMIZATIONS` permission so that users will just get a pop up now asking them to turn off battery optimization instead of having to go dig through their settings as per issue [980](https://github.com/e-mission/e-mission-docs/issues/980). 

This is also an extension of [this closed PR](https://github.com/e-mission/e-mission-data-collection/pull/208). I created this new PR to clean up the commit history as the previous PR was really hard to understand.

PR #211, opened 2023-10-23
Bumped up plugin.xml to 1.8.1

louisg1337:


shankari:
Note to self: this also fixes the previously missed bump to plugin.xml for 
https://github.com/e-mission/e-mission-data-collection/pull/209

PR #213, opened 2024-01-19
Investigate why trip tracking is glitching by adding additional logs

louisg1337:
Add logs to the FSM to figure out why trips don't track or why their tracking is messed up on Android.

### DIFF
--- a/src/android/DataCollectionPlugin.java
+++ b/src/android/DataCollectionPlugin.java
@@ -18,6 +18,7 @@ import android.content.SharedPreferences;
 import android.content.pm.PackageManager;
 import android.os.Build;
 import android.preference.PreferenceManager;
+import android.os.Bundle;
 
 import com.google.android.gms.common.ConnectionResult;
 import com.google.android.gms.location.LocationRequest;
@@ -231,5 +232,29 @@ public class DataCollectionPlugin extends CordovaPlugin {
                 Log.d(cordova.getActivity(), TAG, "Got unsupported request code "+requestCode+ " , ignoring...");
         }
          */
+    }
+
+    @Override
+    public void onStop() {
+      Log.d(cordova.getActivity(), TAG, "onStop() has been called!");
+    }
+
+    @Override
+    public void onDestroy() {
+      Log.d(cordova.getActivity(), TAG, "onDestroy() has been called!");
+    }
+
+    /**
+     * Documentation states this function is...
+     * 
+     * Called when the Activity is being destroyed (e.g. if a plugin calls out to an external
+     * Activity and the OS kills the CordovaActivity in the background). 
+     * 
+     * Perfect for some log statements so that we know if the OS kills an activity it is communicating with 
+     */
+    @Override
+    public Bundle onSaveInstanceState() {
+      Log.d(cordova.getActivity(), TAG, "onSaveInstanceState() has been called, OS killed an activity!");
+      return null;  
     }
 }
--- a/src/android/DataCollectionPlugin.java
+++ b/src/android/DataCollectionPlugin.java
@@ -245,16 +245,11 @@ public class DataCollectionPlugin extends CordovaPlugin {
     }
 
     /**
-     * Documentation states this function is...
-     * 
-     * Called when the Activity is being destroyed (e.g. if a plugin calls out to an external
-     * Activity and the OS kills the CordovaActivity in the background). 
-     * 
-     * Perfect for some log statements so that we know if the OS kills an activity it is communicating with 
+     * This function is called when the Activity is being destroyed. 
      */
     @Override
     public Bundle onSaveInstanceState() {
-      Log.d(cordova.getActivity(), TAG, "onSaveInstanceState() has been called, OS killed an activity!");
+      Log.d(cordova.getActivity(), TAG, "onSaveInstanceState() has been called, OS killed this activity!");
       return null;  
     }
 }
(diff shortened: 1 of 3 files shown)


shankari:
@louisg1337 I will push this to staging once the UI changes have been pushed to production. you should also be able to manually kill the foreground service using `adb`, at least for testing why the foreground service doesn't restart after it is killed

louisg1337:
**Previous Changes**

- Added in foreground service check in `ActivityRecognitionChangeIntentService` and checked to see if location services were available


**New Changes**

- Foreground service check when `DataCollectionPlugin.java / onResume()` gets called already implemented ([here](https://github.com/e-mission/e-mission-data-collection/blob/24d9aa58f141af39631890983a05dee18145300b/src/android/DataCollectionPlugin.java#L68))
- Added in log for when `DataCollectionPlugin.java` calls `onStop()`, `onDestroy()`, and `onSaveInstanceState()`
  - Last one is interesting, apparently it gets called when the plugin is communicating with an Activity that is getting killed by the OS ([docs here](https://github.com/apache/cordova-android/blob/3503bfa31b10cc7d6948356d4c101c7b76109688/framework/src/org/apache/cordova/CordovaPlugin.java#L189)).
- Check for FSM state in `ActivityRecognitionChangeIntentService`

src/android/DataCollectionPlugin.java
```diff
@@ -232,4 +233,28 @@ public void onActivityResult(int requestCode, int resultCode, Intent data) {
...
+     * Documentation states this function is...
+     * 
+     * Called when the Activity is being destroyed (e.g. if a plugin calls out to an external
+     * Activity and the OS kills the CordovaActivity in the background). 
+     * 
+     * Perfect for some log statements so that we know if the OS kills an activity it is communicating with 
+     */
+    @Override
+    public Bundle onSaveInstanceState() {
+      Log.d(cordova.getActivity(), TAG, "onSaveInstanceState() has been called, OS killed an activity!");
```
shankari, line 257: nit: shouldn't this be "OS killed *this* activity"?

src/android/location/ActivityRecognitionChangeIntentService.java
```diff
@@ -50,7 +56,38 @@ protected void onHandleIntent(Intent intent) {
 			Log.i(this, TAG, "Detected new activity "+mostProbableActivity);
 			if (ConfigManager.getConfig(this).isSimulateUserInteraction()) {
 			NotificationHelper.createNotification(this, ACTIVITY_IN_NUMBERS, null, this.getString(R.string.detected_new_activity, activityType2Name(mostProbableActivity.getType(), this)));
+		}
+
+		// Add in logs to check for trip dissapearance
+		Log.d(this, TAG, "Beginning checks for trip dissapearance...");
```
shankari, line 62: nit: spelling

shankari, line 62: nit: spelling

src/android/location/ActivityRecognitionChangeIntentService.java
```diff
@@ -50,7 +56,38 @@ protected void onHandleIntent(Intent intent) {
...
+			// Handle failure, if any
+			Log.d(this, TAG, "Error getting location availability: " + e.getMessage());
+		});
+		
+		// Check for current state
+		SharedPreferences mPrefs = PreferenceManager.getDefaultSharedPreferences(this);
+        String mCurrState = mPrefs.getString(this.getString(R.string.curr_state_key),
+            this.getString(R.string.state_start));
+
+		Log.d(this, TAG, "Ending checks for trip dissapearance...");
```
shankari, line 89: we get this state, but we don't log it anywhere?!

src/android/location/ActivityRecognitionChangeIntentService.java
```diff
@@ -50,7 +56,38 @@ protected void onHandleIntent(Intent intent) {
...
+				Log.d(this, TAG, "Location is available");
+			} else {
+				// Location is not available
+				Log.d(this, TAG, "Location is not available");
 			}
+		})
+		.addOnFailureListener(e -> {
+			// Handle failure, if any
+			Log.d(this, TAG, "Error getting location availability: " + e.getMessage());
+		});
```
shankari, line 82: Instead of checking location availability directly, why not use the methods in the background checks?

shankari (commented):
I am generally supportive of this change, but would like to see the answers ASAP (if possible) so I can merge and push out a release to staging...

src/android/location/ActivityRecognitionChangeIntentService.java
louisg1337: Great point, I think I wrote this part a while ago when I wasn't too familiar with the code base. Converted this to use `checkLocationSettings` in `SensorControlChecks.java`.

src/android/location/ActivityRecognitionChangeIntentService.java
louisg1337: sorry about that, thank you for catching that!

src/android/DataCollectionPlugin.java
louisg1337: Yes I think you're right, I misread the docs sorry about that.

src/android/location/ActivityRecognitionChangeIntentService.java
```diff
@@ -50,7 +60,30 @@ protected void onHandleIntent(Intent intent) {
...
+
+		// Check to see if the foreground service is still alive
+		Context ctxt = getApplicationContext();
+		TripDiaryStateMachineForegroundService.checkForegroundNotification(ctxt);
+
+		// Check if can retrieve current location
+		OnCompleteListener<LocationSettingsResponse> callback = new OnCompleteListener<LocationSettingsResponse>() {
+			@Override // no-op
+			public void onComplete(Task<LocationSettingsResponse> task) { }
+		};
```
shankari, line 76:
```suggestion
			public void onComplete(Task<LocationSettingsResponse> task) {
                           Log.d(this, TAG, "While checking location status in activity response, received callback with : " + task);
			}
		};
```

shankari (approved):

shankari:
@louisg1337 you also need to bump up the version in `plugin.xml`

shankari (approved):

PR #218, opened 2024-03-06
Bluetooth permissions functionality 

louisg1337:
Incorporated new functionality to request bluetooth permissions on Android.

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -214,7 +214,7 @@
     </config-file>
 
     <config-file target="*-Info.plist" parent="NSBluetoothAlwaysUsageDescription">
-      <string>We need Bluetooth access for fleet functionality functionality</string>
+      <string>We need Bluetooth access to interact with BLE beacons for the fleet version of the app.</string>
     </config-file>
 
     <podspec>
--- a/plugin.xml
+++ b/plugin.xml
@@ -57,6 +57,7 @@
         <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
         <uses-permission android:name="android.permission.ACCESS_BACKGROUND_LOCATION"/>
         <uses-permission android:name="android.permission.REQUEST_IGNORE_BATTERY_OPTIMIZATIONS"/>
+        <uses-permission android:name="android.permission.BLUETOOTH_SCAN" />
     </config-file>
 
     <config-file target="AndroidManifest.xml" parent="/manifest/application">
--- a/plugin.xml
+++ b/plugin.xml
@@ -54,10 +54,10 @@
         <uses-permission android:name="android.permission.ACTIVITY_RECOGNITION"/>
         <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
         <!-- COARSE_LOCATION obfuscates the location to a city block, change to FINE_LOCATION for accuracy -->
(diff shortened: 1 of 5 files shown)


shankari (approved):
I don't have any issues with these changes per-se except for the extraneous whitespace. And eventually we do want to incorporate the bluetooth scanning directly into e-mission-datacollection, particularly when we use the callbacks to modify the FSM.

*However* I don't think this should be strictly required - I would assume that the plugin would already have a method to request permissions. Or was it so old that it was written at a time without bluetooth permissions?

Also, where are the iOS changes?

the-bay-kay:
> ... Also, where are the iOS changes?

This is a question I can answer!  From what I understand, the changes were in plugin.xml, line 216:

```xml
    <config-file target="*-Info.plist" parent="NSBluetoothAlwaysUsageDescription">
      <string>We need Bluetooth access to interact with BLE beacons for the fleet version of the app.</string>
    </config-file>
```

This permission was then checked on the UI side, in [this](https://github.com/e-mission/e-mission-phone/blob/1e47df37490ff26424e308864b6263d8a360b3a4/www/js/control/BluetoothScanSettingRow.tsx#L10) component. As we  we understood it, the Classic plugin on iOS only worked with a preset list of peripherals (see screenshot below), and effectively did not scan.  Because this wasn't functional, we opted to leave a "Classic Not Available" error message here!  I can run another test on this branch if need be: the old code for iOS can be found [here](https://github.com/e-mission/e-mission-phone/blob/1b772e2c5217ac8ef80b84cc9413bd808620a320/www/js/control/BluetoothScanSettingRow.tsx#L27).  

ios_bluetooth

louisg1337:
**indicate testing done (showing the new permissions)**

To answer your question about why we wrote our own permissions, the plugin did have a way to request permissions but it was too all over the place. It involved a lot more hassle as the permissions for the plugin requests multiple permissions that we don't need (locations + bluetooth connecting). We also will eventually need to incorporate the permissions for bluetooth so we figured why not just do it now. For testing done, I tested 3 cases on Android only.

I tested what happens when a user, from a fresh install, tries to use the bluetooth scanner. The result is as follows. Clicking allow then takes you to the scanner page.

Screenshot_1710354268

After that, from a fresh install, I tried to deny the permissions. This is the result. Clicking the scanner again will pop up the previous permissions request.

Screenshot_1710354281

Then finally, I tried revoking the permissions to see if it would let you use the scanner, and it still pops up the same permission request as the first picture.

**remove the whitespace**

Just pushed a new commit with the whitespace removed, sorry about that!

**answer the question about iOS**

Katie hit the nail on the head! Because the classic didn't work the way we wanted on iOS, we chose to forego it. We did need that line in the `plugin.xml` to allow us to build or else it causes an error. Along with that, to our knowledge, there is no way to directly ask the user for bluetooth permissions. The way it happens, is iOS requests permissions whenever the `CBCentralManager` is initialized (see [this](https://github.com/pauldemarco/flutter_blue/pull/599) github post). This change had to occur in the bluetooth classic plugin itself, so we made the changes [here](https://github.com/MaximBelov/cordova-plugin-bluetooth-classic-serial-port/commit/71ba673683935c09bcca48bebe218d0fda0a5d46) in this forked version of the plugin. However, since we aren't using the bluetooth classic for iOS we just forewent it and popped up the error like Katie said.

PR #219, opened 2024-03-25
🤖🛜 Android BLE Integration

louisg1337:
This thread will relate to all changes being made to implement BLE functionality into the Android side of 
`e-mission-datacollection`.

This is a continuation of the work done in these following PRs where we created a [Bluetooth Classic](https://github.com/e-mission/e-mission-phone/pull/1128) and [BLE](https://github.com/e-mission/e-mission-phone/pull/1135) scanner.

The core Issue this relates to is [https://github.com/e-mission/e-mission-docs/issues/1046](https://github.com/e-mission/e-mission-docs/issues/1046)

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -40,6 +40,7 @@
   <hook type="after_plugin_install" src="hooks/android/addResourcesClassImport.js" />
 
   <platform name="android">
+    <framework src="src/android/e-mission-datacollection.gradle" custom="true" type="gradleReference" />
 
     <config-file target="res/xml/config.xml" parent="/*">
       <feature name="DataCollection">
--- a/plugin.xml
+++ b/plugin.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <plugin xmlns="http://www.phonegap.com/ns/plugins/1.0"
         id="cordova-plugin-em-datacollection"
-        version="1.8.3">
+        version="1.8.4">
 
   <name>DataCollection</name>
   <description>Background data collection FTW! This is the part that I really
@@ -81,6 +81,8 @@
                 <action android:name="local.transition.stopped_moving"></action>
                 <action android:name="local.transition.stop_tracking"></action>
                 <action android:name="local.transition.start_tracking"></action>
+                <action android:name="local.transition.ble_beacon_found"></action>
+                <action android:name="local.transition.ble_beacon_lost"></action>
                 <action android:name="local.transition.tracking_error"></action>
                 <action android:name="local.transition.checking_for_beacon"></action>
                 <action android:name="local.transition.beacon_found"></action>
@@ -183,6 +185,7 @@
     <source-file src="src/android/wrapper/Transition.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/wrapper"/>
     <source-file src="src/android/wrapper/LocationTrackingConfig.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/wrapper"/>
     <source-file src="src/android/wrapper/Battery.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/wrapper"/>
+    <source-file src="src/android/wrapper/BluetoothBLE.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/wrapper"/>
     <source-file src="src/android/wrapper/ConsentConfig.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/wrapper"/>
     <source-file src="src/android/wrapper/Timer.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/wrapper"/>
     <source-file src="src/android/wrapper/StatsEvent.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/wrapper"/>
@@ -263,6 +266,7 @@
     <header-file src="src/ios/Wrapper/Transition.h" target-dir="Wrapper"/>
     <header-file src="src/ios/Wrapper/LocationTrackingConfig.h" target-dir="Wrapper"/>
     <header-file src="src/ios/Wrapper/Battery.h" target-dir="Wrapper"/>
+    <header-file src="src/ios/Wrapper/BluetoothBLE.h" target-dir="Wrapper"/>
     <header-file src="src/ios/Wrapper/ConsentConfig.h" target-dir="Wrapper"/>
     <header-file src="src/ios/Wrapper/Timer.h" target-dir="Wrapper"/>
     <header-file src="src/ios/Wrapper/StatsEvent.h" target-dir="Wrapper"/>
@@ -285,6 +289,7 @@
     <source-file src="src/ios/Wrapper/Transition.m" target-dir="Wrapper"/>
     <source-file src="src/ios/Wrapper/LocationTrackingConfig.m" target-dir="Wrapper"/>
     <source-file src="src/ios/Wrapper/Battery.m" target-dir="Wrapper"/>
+    <source-file src="src/ios/Wrapper/BluetoothBLE.m" target-dir="Wrapper"/>
     <source-file src="src/ios/Wrapper/ConsentConfig.m" target-dir="Wrapper"/>
     <source-file src="src/ios/Wrapper/Timer.m" target-dir="Wrapper"/>
     <source-file src="src/ios/Wrapper/StatsEvent.m" target-dir="Wrapper"/>
--- a/plugin.xml
+++ b/plugin.xml
@@ -125,6 +125,11 @@
 		    android:enabled="true"
 		    android:exported="false">
 		</service>
+    <service
+		    android:name="edu.berkeley.eecs.emission.cordova.tracker.bluetooth.BluetoothService"
+		    android:enabled="true"
+		    android:exported="false">
+		</service>
     </config-file>
 
     <framework src="com.google.code.gson:gson:2.10.1"/>
@@ -173,6 +178,7 @@
     <source-file src="src/android/wrapper/ConsentConfig.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/wrapper"/>
     <source-file src="src/android/wrapper/Timer.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/wrapper"/>
     <source-file src="src/android/wrapper/StatsEvent.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/wrapper"/>
+    <source-file src="src/android/bluetooth/BluetoothService.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/bluetooth"/>
     <resource-file src="res/android/statemachine.xml" target="res/values/statemachine.xml" />
     <resource-file src="res/android/values/dc_strings.xml" target="res/values/dc_strings.xml"/>
   </platform>
--- a/plugin.xml
+++ b/plugin.xml
@@ -133,6 +133,11 @@
 		    android:enabled="true"
 		    android:exported="false">
 		</service>
+    <service
+		    android:name="edu.berkeley.eecs.emission.cordova.tracker.bluetooth.BluetoothMonitoringService"
+		    android:enabled="true"
+		    android:exported="false">
+		</service>
     </config-file>
 
     <framework src="com.google.code.gson:gson:2.10.1"/>
@@ -182,6 +187,7 @@
     <source-file src="src/android/wrapper/Timer.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/wrapper"/>
     <source-file src="src/android/wrapper/StatsEvent.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/wrapper"/>
     <source-file src="src/android/bluetooth/BluetoothService.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/bluetooth"/>
+    <source-file src="src/android/bluetooth/BluetoothMonitoringService.java" target-dir="src/edu/berkeley/eecs/emission/cordova/tracker/bluetooth"/>
     <resource-file src="res/android/statemachine.xml" target="res/values/statemachine.xml" />
     <resource-file src="res/android/values/dc_strings.xml" target="res/values/dc_strings.xml"/>
   </platform>
--- a/plugin.xml
+++ b/plugin.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <plugin xmlns="http://www.phonegap.com/ns/plugins/1.0"
         id="cordova-plugin-em-datacollection"
-        version="1.8.4">
+        version="1.8.5">
 
   <name>DataCollection</name>
   <description>Background data collection FTW! This is the part that I really
@@ -84,9 +84,6 @@
                 <action android:name="local.transition.ble_beacon_found"></action>
                 <action android:name="local.transition.ble_beacon_lost"></action>
                 <action android:name="local.transition.tracking_error"></action>
-                <action android:name="local.transition.checking_for_beacon"></action>
-                <action android:name="local.transition.beacon_found"></action>
-                <action android:name="local.transition.beacon_not_found"></action>
             </intent-filter>
         </receiver>
         <service
--- a/plugin.xml
+++ b/plugin.xml
@@ -74,7 +74,7 @@
         <receiver
             android:name="edu.berkeley.eecs.emission.cordova.tracker.location.TripDiaryStateMachineReceiver"
             android:enabled="true"
-            android:exported="false">
+            android:exported="true">
             <intent-filter>
                 <action android:name="local.transition.initialize"></action>
                 <action android:name="local.transition.exited_geofence"></action>
@@ -82,6 +82,7 @@
                 <action android:name="local.transition.stop_tracking"></action>
                 <action android:name="local.transition.start_tracking"></action>
                 <action android:name="local.transition.tracking_error"></action>
+                <action android:name="local.transition.checking_for_beacon"></action>
             </intent-filter>
         </receiver>
         <service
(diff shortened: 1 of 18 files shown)


louisg1337:
### **Android Integration Update**
For this BLE integration, a key component is going to be a library called [altbeacon](https://github.com/AltBeacon/android-beacon-library), which will allow us to seamlessly interact with beacons that transmit the iBeacon protocol. The plugin we used for the BLE scanner used this library for Android and it seemed to work great. The library is also actively maintained and very well documented which is a great plus.

With that in mind, I have added altbeacon as a dependency to the Android side of `e-mission-datacollection`. I then roughly implemented a `BluetoothService` class which mimics parts of their [example code](https://altbeacon.github.io/android-beacon-library/samples-java.html) to see if I could scan for my beacon, and I'm happy to say it works! We can receive callbacks in the same way that we could in the BLE scanner, such that we can be notified if a beacon is inside/outside the region.

This is honestly great news as it seems like integrating BLE on Android won't be all the hassle we originally thought it was. 

### **Plan Going Forward**
Everything I have done so far has been a proof of concept, making sure we could 1) download the dependency and 2) use it to scan for a beacon. I think next steps are _roughly_ as follows:

- Explore the [altbeacon API docs](https://altbeacon.github.io/android-beacon-library/javadoc/index.html) a bit more to see what features we can utilize. 
- Clean up `BluetoothService.java` and make it more dynamic
  - Be able to read beacons from a list and start monitoring for each instead of hard coded values.
  - Figure out if extending `Service` is the best option.
    - Explore using `Thread` or `Worker` for optimal performance.
- Explore where in the FSM we will be scanning for the beacons.
  - Figure out how to make the `didDetermineStateForRegion` callback communicate with the FSM (_Intents?_)

JGreenlee:
I think if we move forward with the plan to use one UUID and just differentiate beacons by major/minor (described in https://github.com/e-mission/e-mission-docs/issues/1062), this implementation should switch from the "monitoring" API to the "ranging" API (examples of both in https://altbeacon.github.io/android-beacon-library/samples.html)

JGreenlee:
My understanding is that Monitoring will tell us if you enter or leave a region (and regions are differentiated by UUID).

But if there are multiple beacons in one region, we need Ranging to know when the individual beacons leave or enter. (and we will also get a granular measurement of each beacon's distance)

shankari:
Per the iOS documentation, we should really be using a two-stage process.

First use `Monitoring`, which is much lighter weight and lower power. Once we get the Monitoring callback, we should switch to Ranging, which is more energy intensive.

See the preferred example here:
https://developer.apple.com/documentation/corelocation/determining_the_proximity_to_an_ibeacon_device?language=objc

I anticipate it will be similar on android.

After we get the basic functionality to work, we should experiment with power drain/efficiency by potentially turning off the  Ranging after we have determined the matching beacon.

shankari:
@louisg1337 @JGreenlee before we pivot to the new data model, we should verify that this holds for iOS as well. I have merged https://github.com/e-mission/e-mission-phone/pull/1140 and created a new staging release (1.7.2). @JGreenlee and @louisg1337 can you please update your iPhones and verify that it behaves the same way (don't use the fake callback!)

JGreenlee:
I set both my beacons to the same UUID (I don't know how to change the Bluecharm's UUID so I set the Accent beacon to the Bluecharm UUID)
Then I was able to see both beacons show up when that region was scanned for.



At first I thought it was wrong because I was expecting `1355` for the `minor` of the Bluecharm.
But then I realized it was `1355` in hexadecimal which is `4949` in decimal.

The minor of the Accent beacon I just set to `0001`, or `1`

shankari:
Great! just out of curiosity, I note that one of the beacons had `ProximityNear` and the other had `ProximityFar`. Were they really very far apart? I wonder how sensitive the proximity detection is. Now that we can see these values, we should experiment with this with different phones and beacons.

JGreenlee:
The 'proximity' seems really sensitive to small changes. I can get ProximityImmediate when it's ~6 inches away. ProximityNear from ~1-3 feet away. I get ProximityFar anywhere from ~4-25 feet until it drops off.

It also seems to be significantly affected by some object blocking the direct path from the phone to the beacon (for example, when I put my hand over the beacon).

JGreenlee:
I think we can change the signal strength if we need to

shankari:
Interesting, we should experiment with these in cars and figure out the appropriate combo of signal strength and Proximity to use so that we detect the beacon in the current car, but not the beacon in a car parked nearby. But that is tuning for next month after we have the basic implementation working

louisg1337:
### **FSM Integration Update**
So the past few days I've been wrestling with how to implement this and have been facing lots of setbacks that I will document here. The goal that I am trying to achieve is when we receive the `local.transiiton.exited_geofence` intent, I want to check to see if there are any beacons in range, and if there is, then continue as normal, or if not, restart the FSM. 

In terms of the specifics, I wanted to make it so that we get the intent, we start the scanner, let it scan for 2-3 seconds to ensure we pickup any beacons in the area, and then determine next steps. I wanted to do the 2-3 second wait as I have run into an issue a few times where the scanner does not pick up your beacon on the first or second scan, but does on the third. The little bit of delay would ensure that we will definitely pickup if there is a beacon nearby. 

With that in mind, I figured this code could go into `TripDiaryStateMachineReceiver` as this is the first point of contact that the intents reach when entering the FSM. Here is where the problems start coming in though

1. `BroadcastReceiver` troubles
We have to run the altbeacon code in it's own `Service` as we will continually be getting callbacks from the `MonitorNotifier()` telling us if a beacon is in the region or not. I was trying to figure out how to send data between our `BluetoothService` and the `TripDiaryStateMachineReceiver`, but I'm not quite sure how to do it. You are apparently not allowed to bind a service to `BroadcastReceivers`, so communication via a `Binder` can't work. I thought about sending an intent to `TripDiaryStateMachineReceiver` with the data, but we need the data in our current instance, not another instance of `onReceive`, so that may not be able to work. 

2. Threads are super confusing
In light of everything above, I decided I wanted to try a class based approach instead. My thought process was that with a default class we could access data within it, and most importantly, it would be synchronous which would help with making the code wait 2-3 seconds, as mentioned above. I pasted the code below which shows how I tried to implement this. I created a wrapper class that dealt with data initialization and the methods the program would access. The scanner is defined in an inner class that spawns a separate thread so we can sleep `isBeaconInRange` while simultaneously scanning for beacons. This approach kind of worked, the scanner returned callbacks in the background while the `isBeaconInRange` function slept, but the callbacks never detected a beacon which was super weird. Along with that, I had to keep run() in a while loop as threads apparently quit right when the function ends, which meant we would not receive callbacks. This approach still seems really promising, maybe I am just missing something here that could be the key to fix all of this.

```
    private static String TAG = "BluetoothService";
    private BeaconManager beaconManager;
    public static Context context;


    public BluetoothService(Context context) {
        this.context = context;
    }


    public boolean isBeaconInRange() throws InterruptedException{
        Log.d(context, TAG, "calling isBeaconInRange!");
        beaconManager = BeaconManager.getInstanceForApplication(context);

        // Start the beacon scan in a separate thread
        BeaconScan beaconScan = new BeaconScan(context);
        beaconScan.start();

        // Wait for 5 seconds to let the beacon potentially come in range
        Thread.sleep(5000);

        // Stop it now
        beaconScan.stopBeaconScan();

        Log.d(context, TAG, "just slept 5 seconds, value in beaconScan is...");
        boolean inRange = beaconScan.getBeaconInRange();
        Log.d(context, TAG, "inRange: " + inRange);

        return inRange; 
    }

    public class BeaconScan extends Thread {
        private volatile boolean beaconInRange;
        public Context context;
        private boolean keepScanning = true;
        private boolean ranOnce = false;

        public BeaconScan(Context context){
            this.context = context;
        }

        @Override
        public void run() {
            while(keepScanning) {
                // Code to start scanning for BLE beacons using AltBeacon library
                Log.d(context, TAG, "startBeaconScan called!!!!");

                beaconManager.addMonitorNotifier(new MonitorNotifier() {
                    @Override
                    public void didEnterRegion(Region region) {
                        Log.d(BluetoothService.context, TAG, "Beacon entered the region. Beacon UID: " + region.getUniqueId());
                    }
            
                    @Override
                    public void didExitRegion(Region region) {
                        Log.d(BluetoothService.context, TAG, "Beacon has exited the region. Beacon UID: " + region.getUniqueId());
                    }
            
                    @Override
                    public void didDetermineStateForRegion(int state, Region region) {
                        // 0 = OUTSIDE || 1 = INSIDE
                        
                        if (state == 0) {
                            Log.d(BluetoothService.context, TAG, "Beacon is not inside the region! Beacon UID: " + region.getUniqueId());
                        } else {
                            Log.d(BluetoothService.context, TAG, "Beacon is inside the region! Beacon UID: " + region.getUniqueId());
                            Log.d(BluetoothService.context, TAG, region.toString());
                            beaconInRange = true;
                        }
                    }
                });

                if (!ranOnce) {
                    beaconManager.startMonitoring(new Region("426C7565-4368-6172-6D42-6561636F6E73", null, null, null));
                }
                ranOnce = true;

                // Sleep for a short duration of time
                try {
                    Thread.sleep(500); 
                } catch (InterruptedException e) {
                    e.printStackTrace();
                }
            }
        }
    
        public void stopBeaconScan() {
            Log.d(context, TAG, "Stopping monitoring for the beacon.");
            beaconManager.stopMonitoring(new Region("426C7565-4368-6172-6D42-6561636F6E73", null, null, null));
            beaconManager.removeAllMonitorNotifiers();
            keepScanning = false;
        }

        public boolean getBeaconInRange() {
            return beaconInRange;
        }
   
    }

}


```

3. Make the changes in `TripDiaryStateMachineService` instead?
As I haven't found anything that has worked in `TripDiaryStateMachineReceiver`, I figured the second best place to make changes in is `TripDiaryStateMachineService` as that is the next place the FSM goes to after the receiver. It also is a `Service`, not a `BroadcastReceiver`, so I figured I could use `Binders` and make the bluetooth scanner a service again. I have been trying to implement that for a little while now but the app keeps on crashing on me. A big issue that I am facing is that there is a decent delay on `bindService`. If I try to immediately access methods in my `BluetoothService` I get a null pointer exception as I am accessing it too early, and the bind has yet to be completed. I still have a bit of hope left for this method though, I do want to play around with it a bit more and see if I can get it to work. 

I would definitely appreciate any ideas or thoughts on a better way to implement this. I feel like there definitely is a clean way to accomplish all of this, I just don't know it yet.

shankari:
@louisg1337 couple of high level comments/suggestions:
- I am not sure you have learned about FSM (Finite State Machines) in school. Here we are using the FSM as a way to clarify and code the complexities of the tracking. An obvious solution to the goal you have here is to add a new state (CHECKING_FOR_BEACON) to the FSM for the fleet case. See the image below for what the fleet versus non-fleet option may look like

Screenshot 2024-03-31 at 9 19 56 AM

Now that you see it visually, you can also see the holes in the plan. We want to spend most of our time in "Waiting for trip start"; once we have moved to "Checking for beacon", how do we get back to "Waiting for trip start" if it is not a fleet trip. The answer is obvious (timeout), but I think it would be helpful for you to draw out what the FSM looks like for both the location-triggered and beacon-triggered cases.

You can see the existing services (e.g. `LocationChangeIntentService.java` or `GeofenceExitIntentService.java`) for examples of how to send messages to the FSM. I hope you have also checked the original FSM from Figure 4.5 of my thesis for the baseline.

- If you do still need to have a separate thread in the `TripDiaryStateMachineService`, you can look at how we handle Geofence creation, which needs to run in a separate thread (aka `createGeofenceInThread`). But I would suggest adding new states to the FSM first, since it is much more clear

shankari:
> I'll take some time in the next few days to dig into the thesis and FSM diagram some more to hopefully answer the questions above.

Yes. You are close. High level comments:
- The `TripDiaryStateMachineReceiver` is typically a pass-through to `TripDiaryStateMachineService`. The FSM is implemented in `TripDiaryStateMachineService`. There is basically one function for each state, and within the function, we check the current state and the transition, perform the required actions, and then move to a new state.
- So the steps above would actually be:
   - We receive `local.transition.exited_geofence` in `TripDiaryStateMachineReceiver`
   - We send it to `TripDiaryStateMachineService`
   - `TripDiaryStateMachineService` `handleAction` calls `handleTripStart` since the trip has not yet started
   - `handleTripStart` sees that we are a fleet deployment, so instead of starting location tracking, it starts the beacon read. After the beacon reading service has started, the FSM moves to `checking_for_beacon` state
   - if a beacon is found, the beacon service sends a `found_beacon` message to the receiver, which sends it to the server, which calls a new `handleBeaconFound` state.
   - since this is a fleet deployment, `handleBeaconFound` deletes the geofence, starts trip tracking etc

Again, I think that understand the current FSM patterns and drawing out your new plan as an FSM will help you a lot

shankari:
> The other update is that I realized I wasn't considering something important, which was running the BluetoothService in the background when the app is closed. So far, everything works when the app is in the foreground, but once you put it in the background, BluetoothService gets called, but no beacon ranging occurs.

I thought that this shouldn't matter. As you may remember from our deep dive into the location tracking, we start a foreground service for the app, and when that is running, it should be equivalent to the service starting in the foreground. However, on reading up on foreground services, I see the following
https://developer.android.com/develop/background-work/services/foreground-services#start

> For example, suppose a fitness app runs a running-tracker service that always needs location information, but might or might not need to play media. You would need to declare both location and mediaPlayback in the manifest. If a user starts a run and just wants their location tracked, your app should call startForeground() and pass just the ACCESS_FINE_LOCATION permission. Then, if the user wants to start playing audio, call startForeground() again and pass the bitwise combination of all the foreground service types (in this case, ACCESS_FINE_LOCATION|FOREGROUND_SERVICE_MEDIA_PLAYBACK).

I am not quite sure if the bluetooth service needs to be in the foreground or you just need to bind to the existing foreground service. I would recommend the second option so that we don't pollute the user's notification bar with too many foreground services. I would also suggest that you experiment with just starting foreground with location and `connectedDevice` and see if it works.

louisg1337:
> I am not quite sure if the bluetooth service needs to be in the foreground or you just need to bind to the existing foreground service. I would recommend the second option so that we don't pollute the user's notification bar with too many foreground services. I would also suggest that you experiment with just starting foreground with location and connectedDevice and see if it works.

I totally agree, I don't think starting another foreground service is the right thing to do, but thankfully though, it seems like we don't need to. I just pushed a commit that lets us scan beacons in the background, and all I am doing now is starting `BluetoothService` from `TripDiaryStateMachineForegroundService`. Let me know what you think about this solution @shankari , if not I can try to do the binding version if you think it would be better.

JGreenlee:
> The next thing I wanted to mention was that for the time being, to check if we are using a fleet config, I just did
> `if (true)`. I'm just using this as a placeholder until we get the config setup.

I did get around to testing this. Long story short – the config can be retrieved fairly easily by using the `getDocument()` function from the UserCache plugin.
```java
String app_config_key = ctxt.getString(R.string.key_usercache_app_config);
JSONObject config = (JSONObject) UserCacheFactory.getUserCache(ctxt).getDocument(app_config_key, false);
```
(This depends on a new key that https://github.com/e-mission/cordova-usercache/pull/50 adds. Until that is merged, you could just hardcode `"config/app_ui_config"`)

<hr>

The way I tested this was by adding a new Cordova JS function `getAppConfig` which retrieves the config in native code and returns it over the Cordova bridge (so I can easily see that is working)

### DataCollectionPlugin.java
```java
  ...
  } else if (action.equals("getAppConfig")) {
      try {
          Log.d(cordova.getActivity(), TAG, "getting app config");
          Context ctxt = cordova.getActivity();
          String app_config_key = ctxt.getString(R.string.key_usercache_app_config);
          JSONObject config = (JSONObject) UserCacheFactory.getUserCache(ctxt).getDocument(app_config_key, false);
          callbackContext.success(config);
          return true;
      } catch (Exception e) {
          Log.e(cordova.getActivity(), TAG, "Error getting app config");
          callbackContext.error(e.getMessage());
          return false;
      }
  }
  ...
```

### datacollection.js
```js
...
getAppConfig: function() {
    return new Promise(function(resolve, reject) {
        exec(resolve, reject, "DataCollection", "getAppConfig", []);
    })
}
...
```

Then a new row in the Profile tab to display the retrieved config:
### ProfileSettings.tsx
```jsx
...
<SettingRow
  textKey="Read app config from native"
  iconName="application"
  action={() => {
    window['cordova'].plugins.BEMDataCollection.getAppConfig().then((result) => {
      const cfg = JSON.stringify(result, null, 2);
      logInfo('appConfig = ' + cfg);
      window.alert('appConfig = ' + cfg);
    });
  }}
/>
...
```

<hr>

### It works!

louisg1337:
### **Newest Commit Update**

I just pushed a new commit that has a few new changes that I'll explain below.

**Config integration**
I managed to get the config loaded in and it now checks for `"tracking": { "bluetooth_only": true }` as noted [here](https://github.com/e-mission/e-mission-docs/issues/1062#issuecomment-2038046668). As long as that stays the same, and we are still using `"config/app_ui_config"` to access it, we should be good. Big thanks to @JGreenlee for doing the heavy lifting on helping us figure out how to access the config from the native code.

**Permissions**
I added a bluetooth permission check to `SensorControlChecks`, and then called that in our `BluetoothService` to ensure that we only range if we have permissions. I think it won't cause any errors if we try to range without permissions, but this is just to be safe.

**Bluetooth Monitoring**
I attempted to add in a `BluetoothMonitoringService` which would continually monitor for region exits/enters upon geofence creation. However, I commented out the code that would start up the service as its a bit buggy right now. First of all, monitoring doesn't seem to work, as it won't give me updates whenever I turn my beacon on and off again. Second, apparently we can't monitor and range at the same time, as whenever I try to range for beacons while monitoring, ranging doesn't work (in both foreground + backgorund). I'll need to investigate this some more, but for right now I commented it out so it doesn't interfere with the ranging. 

**Testing**
I've been taking my test phone + beacon to and from classes, and it seems like the tracking aspect works. It tracks when I have my beacon, and it doesn't when I don't. Katie had an issue where it still tracked on her test phone despite no beacon, but I think we sorted that out. The issue was that we weren't reinstantiating `private Set<Beacon> scanned;` each time we ranged for beacons, so sometimes the values would persist and the code would think there was a beacon in range.

PR #223, opened 2024-04-15
fix string equality check

JGreenlee:
Should compare strings with equals instead of ==


shankari:
Gah! Too much python and javascript code recently! I remembered to fix it in most cases, but I missed this one because it is a static check. IIRC, `==` does work in java for string comparisons if both sides are static strings (due to some memoization magic) but not here.

Great job!

shankari:
Merged

PR #224, opened 2024-04-15
👷🔧 🪠 Plumb through support for requesting bluetooth scan permissions …

shankari:
…on android

As far as we can tell, iOS does not need any additional permissions to use the iBeacon API since we already have "always" location permissions. However, android requires the BLUETOOTH_SCAN permission https://developer.android.com/develop/connectivity/bluetooth/bt-permissions#declare-android12-or-higher

This change includes:
- removing the hack popup that shows up when the app is launched
- adding check and checkAndFix methods in the javascript -> plugin -> foreground checker
- adding a new check method to the list of checks
- invoking the method from the background checker so that we can prompt people to provide it if they don't already do so
- ensure that the background checker method is only run for fleet configs

Testing done:
- See screenshots in the related UI PR
- Without the fleet check, simulated trip start/end on a non-fleet config and got a notification saying that permissions were incorrect
- With the fleet check, retried, and did not get the notification

### DIFF
--- a/src/android/DataCollectionPlugin.java
+++ b/src/android/DataCollectionPlugin.java
@@ -62,10 +62,6 @@ public class DataCollectionPlugin extends CordovaPlugin {
                 new StatsEvent(myActivity, R.string.app_launched));
 
         TripDiaryStateMachineReceiver.initOnUpgrade(myActivity);
-
-        // Ask for bluetooth permissions
-        // We will change this with future releases, we just ran out of time implementing this into the front end
-        mControlDelegate.checkAndPromptBluetoothScanPermissions();
     }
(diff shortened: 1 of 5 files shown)


shankari:
| Bluetooth permission in fleet mode | Popup | turns green |
|------|--------|-------|
| Screenshot 2024-04-14 at 10 53 08 PM | Screenshot 2024-04-14 at 10 53 41 PM | Screenshot 2024-04-14 at 10 53 56 PM |

| Does not show up in non-fleet mode | or on iOS |
|--------|-------|
| Screenshot 2024-04-14 at 10 59 57 PM |   |

PR #226, opened 2024-04-15
🔵🦷 Initial, proof of concept, alpha version of the bluetooth integration

shankari:
Includes:
- basic monitoring and ranging on iOS
- geofence-triggered ranging on android
- FSM changes to, in the fleet version, only read location in proximity to BLE
- prompting for bluetooth permissions
- wrapper class for bluetooth
- a javascript function to mock BLE objects and aid in testing

Related usercache change:
https://github.com/e-mission/cordova-usercache/pull/51

Related UI changes:
https://github.com/e-mission/e-mission-phone/pull/1144

### DIFF
--- a/hooks/android/addResourcesClassImport.js
+++ b/hooks/android/addResourcesClassImport.js
@@ -77,6 +77,12 @@ module.exports = function (ctx) {
                             reject();
                         }
 
+                        /*
+                        console.log("For file "+file
+                                        + " Results of match checks: R = "+contents.match(R_RE)
+                                        + " BUILDCONFIG = "+contents.match(BUILDCONFIG_RE)
+                                        + " MAINACT = "+contents.match(MAINACT_RE));
+                        */
                         if (contents.match(R_RE) || contents.match(BUILDCONFIG_RE) || contents.match(MAINACT_RE)) {
                             console.log('file '+filename+' needs to be rewritten, checking package');
                             const packages = contents.match(PACKAGE_RE);
@@ -136,6 +142,7 @@ module.exports = function (ctx) {
                             // since we re-check for the imports before re-writing them
                             // but it avoid unnecessary file rewrites, so we retain
                             // it for now
+                            console.log("file "+file+" does not need to be rewritten");
                             resolve();
                         }
                     });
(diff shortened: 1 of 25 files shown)


shankari:
This works, and has been deployed in the alpha version. It passes all the CI checks. *However*, when trying to use it with in a package that is not emission, we run into an issue where the `R` package is not handled properly in the hook.

```
$ npx cordova plugin rm cordova-plugin-em-datacollection
Uninstalling cordova-plugin-em-datacollection from android
Subproject Path: CordovaLib
Subproject Path: app
Uninstalling cordova-plugin-em-datacollection from ios
Running command: pod install --verbose
Removing "cordova-plugin-em-datacollection"
...
...platforms/android/app/src/main/java/edu/berkeley/eecs/emission/cordova/tracker/bluetooth/BluetoothMonitoringService.java:15: error: cannot find symbol
import edu.berkeley.eecs.emission.R;
                                 ^
  symbol:   class R
  location: package edu.berkeley.eecs.emission
...platforms/android/app/src/main/java/edu/berkeley/eecs/emission/cordova/tracker/location/TripDiaryStateMachineForegroundService.java uses or overrides a deprecated API.
Note: Recompile with -Xlint:deprecation for details.
1 error
```

Note that this is the only java file which is affected. We need to see why this is not handled the same way as the others.

shankari:
This is because we don't check for the import statement, we check for the **usage** of the package.

```
const R_RE = /[^\.\w]R\./;
> "import edu.berkeley.eecs.emission.R;".match(R_RE);
null
> "import edu.berkeley.eecs.emission.R;".match(/emission/);
[
  'emission',
  index: 25,
  input: 'import edu.berkeley.eecs.emission.R;',
  groups: undefined
]
> " R.string.transition_ble_beacon".match(R_RE);
[
  ' R.',
  index: 0,
  input: ' R.string.transition_ble_beacon',
  groups: undefined
]
```

So the problem is that we are not using the variable anywhere. Let's add a simple statement to use it...

shankari:
After adding code to save the entries

```
...
                    UserCacheFactory.getUserCache(BluetoothMonitoringService.this)
                        .putSensorData(R.string.key_usercache_bluetooth_ble, currWrapper);
...
```


we get 

```
For file .../plugins/cordova-plugin-em-datacollection/src/android/bluetooth/BluetoothMonitoringService.java Results of match checks: R = (R. BUILDCONFIG = null MAINACT = null
file BluetoothMonitoringService.java needs to be rewritten, checking package
```

PR #227, opened 2024-04-15
Slightly refactored process for how we range beacons

louisg1337:
Changed number of scans we do and made it so we exit early if we found OpenPATH beacon.


shankari:
@louisg1337 can you indicate the testing done? Alternatively, make only the one-line change. If the change is untested, it should be minimal.

PR #228, opened 2024-04-16
📤 📉 Fixed extra `didRangeBeacons` transitions

the-bay-kay:
`didRangeBeacons` will no longer send transition notifications during an ongoing trip.

This change should improve both performance and Native Log Legibility.

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <plugin xmlns="http://www.phonegap.com/ns/plugins/1.0"
         id="cordova-plugin-em-datacollection"
-        version="1.8.5">
+        version="1.8.6">
 
   <name>DataCollection</name>
   <description>Background data collection FTW! This is the part that I really
--- a/src/ios/Location/TripDiaryDelegate.m
+++ b/src/ios/Location/TripDiaryDelegate.m
@@ -358,40 +358,10 @@
                                 showUI:TRUE];
 }
 
-/* In testing, it appears that relying on ExitRegion alone is not enough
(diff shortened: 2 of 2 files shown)


shankari:
@the-bay-kay can you please indicate the testing done? Also, please remember to put this into the "Ready for review" column when you are done.

shankari:
@the-bay-kay still waiting for testing done. You also need to bump up the version number in `plugin.xml`, and after I create the release, in the `package.cordovabuild.json`

the-bay-kay:
@shankari This builds and runs successfully! I can open a PR to bump `package.cordovabuild.json` later today.

I've got a few notes / questions:
- Do we want to log the beacon data each time we range?  This is what I'm doing as written.  This will cause quite a few logs during the trip, but that information could be useful for debugging later. 
- With both this build and the previous (master) build, I noticed that `exitRegion` is not immediately triggering the `CFCBeaconLost`, and subsequently, ending the trip.  This could explain some of the unusual tracking behavior seen in some of the trips (located in the internal github repo).  Since that's a bit more of an involved change, I'll be doing that in a different PR.

shankari:
> Do we want to log the beacon data each time we range? This is what I'm doing as written. This will cause quite a few logs during the trip, but that information could be useful for debugging later.

Yes. We can potentially reduce log spew once we get out of beta

> With both this build and the previous (master) build, I noticed that exitRegion is not immediately triggering the CFCBeaconLost, and subsequently, ending the trip. This could explain some of the unusual tracking behavior seen in some of the trips (located in the internal github repo). Since that's a bit more of an involved change, I'll be doing that in a different PR.

Sounds good.

PR #230, opened 2024-04-30
Fix ios range transition

shankari:


shankari:
@the-bay-kay created new release; you should be able to bump up `package.cordovabuild.json`

PR #231, opened 2024-05-16
Work profile fix

louisg1337:
This PR addresses [this issue](https://github.com/e-mission/e-mission-docs/issues/1070).
- Adds in an additional check to see if app is installed on a work profile
- Reverts previous temporary fix 
- Bumps version number

**Testing Done**
I successfully went through the permissions process for the app while installed on both a work profile and personal profile.

### DIFF
--- a/src/android/verification/SensorControlChecks.java
+++ b/src/android/verification/SensorControlChecks.java
@@ -8,6 +8,8 @@ import android.location.Location;
 import android.os.Build;
 import android.os.PowerManager;
 import android.content.pm.PackageManager;
+import android.app.admin.DevicePolicyManager;
+import android.content.ComponentName;
 
 import androidx.annotation.RequiresApi;
 import androidx.core.app.NotificationManagerCompat;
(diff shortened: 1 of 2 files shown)


src/android/verification/SensorControlChecks.java
```diff
@@ -113,15 +115,43 @@ public static boolean checkNotificationsUnpaused(final Context ctxt) {
...
+  
+  /**
+   * Check if the app is installed on a work profile.
+   * 
+   * We need this check as Android automatically exempts apps installed on a workprofile from hibernation. 
+   * This means that the "Pause app activity if unused" option is greyed out, which blocks users from continuing
+   * past the permissions screen. 
+   */
+  public static boolean checkWorkProfile(final Context ctxt) {
+    DevicePolicyManager devicePolicyManager = (DevicePolicyManager) ctxt.getSystemService(ctxt.DEVICE_POLICY_SERVICE);
```
shankari, line 127: I am fine with merging this as-is so that we don't over-engineer this, but in the future, we may want to/need to return the admin information. But it is fine for now, since it follows the format of the other checks as well.

shankari (approved):

PR #235, opened 2024-09-19
Patch holes that cause unwanted trips to be started in fleet/non-fleet modes.

catarial:
1. Turning off duty cycling in fleet mode causes tracking to be turned on regardless of if there's a beacon or not.
2. Receiving a beacon found transition in non-fleet mode causes a trip to be started.

# Fix

1. Don't start tracking when duty cycling is turned off in fleet mode until we see a beacon.
2. Log this as an error and stay in the same state.

I wasn't able to test if tracking works in fleet mode when duty cycling is turned off since I don't have a beacon. But I was able to verify that it doesn't immediately go into an ongoing trip when duty cycling is turned of in fleet mode.

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <plugin xmlns="http://www.phonegap.com/ns/plugins/1.0"
         id="cordova-plugin-em-datacollection"
-        version="1.9.0">
+        version="1.9.1">
 
   <name>DataCollection</name>
   <description>Background data collection FTW! This is the part that I really
--- a/src/android/location/TripDiaryStateMachineService.java
+++ b/src/android/location/TripDiaryStateMachineService.java
@@ -281,7 +281,8 @@ public class TripDiaryStateMachineService extends Service {
                 if (isFleet) {
                     Log.d(this, TAG, "Found beacon in fleet mode, starting location tracking");
                 } else {
-                    Log.e(this, TAG, "Found beacon in non-fleet mode, not sure why this happened, starting location tracking anyway");
+                    Log.e(this, TAG, "ERROR: Found beacon in non-fleet mode, not starting location tracking");
+                    return;
                 }
             }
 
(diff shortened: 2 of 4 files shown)


shankari (approved):
This is correct, but I would need to look through the FSM carefully to determine if it is complete. In particular, is this the only instance of `oneTimeInitTracking` in the initializer on iOS? I thought that there was one related to `ongoing_trip` as well

Ah yes!

```
    if (self.currState == kOngoingTripState) {
        // TODO: Add restart for BLE Version

        /* StandardVERSION:*/

        // If we restarted, we recreate the location manager, but then it won't have
        // the fine location turned on, since that is not carried through over restarts.
        // So let's restart the tracking
        [TripDiaryActions startTracking:CFCTransitionTripRestarted withLocationMgr:self.locMgr];
        // Note that if the phone was shut down when the app was in the ongoing trip state, and it was
        // turned back on at home, we will start tracking here but will most probably not get a visit transition
        // so the data collection will be turned on until the NEXT trip ends. This is why we need remote pushes, I think.
        // would be good to test, though.
    }
```

But we can deal with that in the next PR.

shankari:
This will happen when the app is killed and then restarted.

a concrete use case that can drive the decision is: user is in a car, phone dies, phone is plugged in again and restarts
- if the phone restarts when the user is in the car, we want to continue tracking
- if the phone restarts when the user is not in the car, we want to stop tracking and go into the "waiting_for_beacon" state

for the non-fleet case, we just start tracking even if the user has got home and has plugged in their phone, we want to know that they got home so that the next trip will start at the correct location. We want to re-establish the geofence etc so we can just start tracking again and let the trip end detection take over to do all that.

For the fleet case, we don't want to read the location if the user has gone home and plugged in there.
So we will need to check for beacon visibility before we restart the trip.

shankari:
So to work with beacons in iOS, you would use the corelocation closed source API
https://developer.apple.com/documentation/corelocation?language=objc

Beacon ranging: Detect and locate nearby beacons.

PR #236, opened 2024-10-03
Stop tracking when duty cycling is turned off in fleet mode

catarial:
We don't want duty cycling to be turned off in fleet mode because it would track trips in non-fleet vehicles. Having duty cycling off and still only trying to track trips in fleet vehicles doesn't make sense because tracking just fleet vehicles is a form of duty cycling.

This patch treats this event as an error rather than an intentional decision.

If we plan on removing this from the UI in fleet mode it might be a good idea to have it automatically try to turn duty cycling back on.

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <plugin xmlns="http://www.phonegap.com/ns/plugins/1.0"
         id="cordova-plugin-em-datacollection"
-        version="1.9.1">
+        version="1.9.2">
 
   <name>DataCollection</name>
   <description>Background data collection FTW! This is the part that I really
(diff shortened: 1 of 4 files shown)


shankari:
I don't think this has been tested sufficiently.

On a fleet deployment (DFC Fermata),

Without this change, on iOS, the behavior is as expected:
`WAITING_FOR_TRIP_START` -> Turn off duty cycling -> `STATE_START` -> Turn on duty cycling -> `WAITING_FOR_TRIP_START`

With this change, on iOS, the behavior is unexpected
`WAITING_FOR_TRIP_START` -> Turn off duty cycling -> `STATE_TRACKING_STOPPED` -> Turn on duty cycling -> `STATE_TRACKING_STOPPED`

due to

```
in restartCollection, state is already TRACKING_STOPPED, early return
DEBUG: in restartCollection, state is already TRACKING_STOPPED, early return
DEBUG:parseState: state = STATE_TRACKING_STOPPED; 
      platformId = ios
```

shankari:
I'm going to move this off our plate for now until we have a chance to think through the FSM more carefully.

PR #237, opened 2024-10-04
🍎 iOS Permissions Changes to Comply with iOS Guidelines

JGreenlee:
As described in https://github.com/e-mission/e-mission-docs/issues/1094

- implements the "2-step strategy" for iOS 13.4+. If the user denies or gives insufficient permissions at any point in the process, subsequent attempts to fix the permission will just take them to the app settings page
- if user notifications permissions get denied, prevents app settings from immediately opening
- updates text for several permissions

### DIFF
--- a/plugin.xml
+++ b/plugin.xml
@@ -212,19 +212,19 @@
     </config-file>
 
     <config-file target="*-Info.plist" parent="NSLocationAlwaysAndWhenInUseUsageDescription">
-        <string>OpenPATH requires "Precise" location access allowed "Always" to create an automatic trip diary</string>
+        <string>To create an automatic trip diary, "Precise" location access must be "Always" allowed</string>
     </config-file>
 
     <config-file target="*-Info.plist" parent="NSLocationAlwaysUsageDescription">
-        <string>OpenPATH also requires "Always" location access to track your trips in the background</string>
+        <string>To track your trips in the background, location access must be "Always" allowed</string>
     </config-file>
 
     <config-file target="*-Info.plist" parent="NSLocationWhenInUseUsageDescription">
-        <string>OpenPATH requires "Precise" location access allowed "While Using App" to create an automatic trip diary</string>
+        <string>To create an automatic trip diary, "Precise" location access must be allowed "While Using App"</string>
     </config-file>
 
     <config-file target="*-Info.plist" parent="NSMotionUsageDescription">
-        <string>OpenPATH uses your phone's motion sensors to determine the modes of transportation during your trips</string>
+        <string>To determine the modes of transportation during your trips, the app must access your phone's motion sensors</string>
     </config-file>
 
     <config-file target="*-Info.plist" parent="ITSAppUsesNonExemptEncryption">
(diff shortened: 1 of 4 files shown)


src/ios/Verification/SensorControlForegroundDelegate.m
```diff
@@ -282,25 +299,39 @@ - (void) didRegisterUserNotificationSettings:(UIUserNotificationSettings*)newSet
...
+    } else if (IsAtLeastiOSVersion(@"13.0")) {
+        NSLog(@"iOS >=13,<13.4 detected, launching UI settings to manually enable 'always'");
         // we want to leave the registration in the prompt for permission, since we don't want to register callbacks when we open the app settings for other reasons
         [[TripDiaryStateMachine delegate] registerForegroundDelegate:self];
         [[TripDiaryStateMachine instance].locMgr startUpdatingLocation];
         [self openAppSettings];
-    }
-    else {
+    } else {
+        NSLog(@"iOS <13 detected, simply requesting 'always'");
```
JGreenlee, line 329: I believe we do not support iOS <13 anymore, so maybe this should be removed

JGreenlee, line 329: I believe we do not support iOS <13 anymore, so maybe this should be removed

shankari: Yup, we can remove that in a cleanup change

plugin.xml
```diff
@@ -212,27 +212,27 @@
     </config-file>
 
     <config-file target="*-Info.plist" parent="NSLocationAlwaysAndWhenInUseUsageDescription">
-        <string>We use your data to create an automatic trip diary</string>
+        <string>OpenPATH requires "Precise" location access allowed "Always" to create an automatic trip diary</string>
```
shankari, line 215: I had this as "We" so that community apps such as Ma Mobilitie and FourStep wouldn't have to edit the plugin to work. Let's revert these and figure out the modified text in yet another cleanup change..

res/android/values/dc_strings.xml
```diff
@@ -17,9 +17,8 @@
     <!-- TripDiaryStateMachineService and Ongoing-->
     <string name="success_moving_new_state">Success moving to %1$s</string>
     <string name="failed_moving_new_state">Failed moving to %1$s</string>
-    <string name="location_permission_off">Insufficient location permissions, please fix</string>
-    <string name="location_permission_off_app_open">Insufficient location permissions, please fix in app settings</string>
-    <string name="location_permission_off_enable">Location permission off, click to enable</string>
+    <string name="location_permission_off">Insufficient permissions, please allow 'Always' and 'Precise' location access</string>
```
shankari, line 20: I had left this vague because it felt like iOS kept adding new permissions and I didn't want to keep changing the plugin. But they haven't changed it for a while, and there are others who now have confidence in changing the native code, so this is fine.

shankari (changes requested):
LGTM once we revert the "OpenPATH" changes.

JGreenlee:
I edited the descriptions to make them not specific to OpenPATH.
I do think it is important to hint to the user at each step what specific type of location access is required ("Precise, "While Using App", "Always Allow") because if the user doesn't do it perfectly the first time (for example if they tap "Just Once" instead of "While Using App") they will have to go back and fix it manually.

shankari (approved):

PR #238, opened 2024-10-07
🔥 Single-line changes to remove notification checks in the background checker

shankari:
As part of https://github.com/e-mission/e-mission-docs/issues/1094, we made the notification permission optional.

Since it is not required for silent push notifications, notifications are nice-to-have. They are still helpful to ensure that people know if there are issues, and hopefully help them label their trips, but the app will still work if the permission is turned off.

Since this is no longer required for normal operation, we don't need to check it in the background every hour, and nag the user to fix it if it is turned off. This is a super-easy change (two lines), so rolling this in while fixing https://github.com/e-mission/e-mission-docs/issues/1094 instead of coming in as a future fix.

Bonus fix: Use the ExplicitIntent wrapper in the geofence exit intent service to be consistent with all the other locations.

### DIFF
--- a/src/android/location/actions/GeofenceLocationIntentService.java
+++ b/src/android/location/actions/GeofenceLocationIntentService.java
@@ -111,8 +111,7 @@ public class GeofenceLocationIntentService extends IntentService {
     }
 
     private void broadcastLoc(Location loc) {
-        Intent answerIntent = new Intent(INTENT_NAME);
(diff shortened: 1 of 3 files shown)


shankari:
@JGreenlee if you think this looks good, I can pull this, your changes, bump up the release and get a new version on to staging ASAP

JGreenlee (approved):
LGTM

PR #240, opened 2024-10-08
store "HasRequestedNotificationPermission" in native LocalStorage

JGreenlee:


shankari:
@JGreenlee this now compiles on both android and iOS (see my last commit).
Once the phone changes are done, I can pull and test...

JGreenlee:
I put all the bugfixes in https://github.com/e-mission/e-mission-phone/pull/1183

PR #242, opened 2024-10-09
Hack to set the `HasRequestedNotificationPermission`

shankari:
If it is not set and the user has consented
This fixes
https://github.com/e-mission/e-mission-docs/issues/1094#issuecomment-2403052620


shankari:
@JGreenlee I am almost out of battery; but you can take a look at this

JGreenlee (approved):
LGTM

PR #244, opened 2025-02-24
Increase logs to see when iOS is killing the app

Abby-Wheelis:
Based on lots of logs indicating that the iOS native code is being initialized, we think the OS is killing the app, increasing logs will help us figure out when this is happening and what is causing it. See [issue #1114](https://github.com/e-mission/e-mission-docs/issues/1114) for further details


Abby-Wheelis:
As it turns out, this was the wrong place for the metricKit implementation, so just adding the memory warning log here

shankari (approved):

PR #246, opened 2025-03-21
👔 Listen to beacons only for fleets

shankari:
Before this change, on android, we used to scan for beacons only for fleet deployments, but on iOS, we used to start scanning for beacons immediately on starting tracking, since the beacons are a "geofence-like" mechanism on iOS.

However, since we only do BLE matching in fleet mode, we can turn off scanning for non-fleet deployments, avoiding wasted working and reducing battery drain.

- This PR also includes refactoring of the codebase to pull out `isFleet` into a separate function that can be reused instead of copy-pasting the code in multiple places.
- As a bonus, it also refactors the consent code so that we read it once and then use the cached copy instead of going to the database every time we need to check the consent

### DIFF
--- a/src/ios/Location/TripDiaryActions.m
+++ b/src/ios/Location/TripDiaryActions.m
@@ -51,6 +51,11 @@ static NSString* const GEOFENCE_LOC_KEY = @"CURR_GEOFENCE_LOCATION";
 }
 
 + (void)startBLEMonitoring:(NSString*) transition withLocationMgr:(CLLocationManager*)locMgr {
+    if (![ConfigManager isFleet]) {
+        [LocalNotificationManager addNotification:
+         [NSString stringWithFormat:@"Not a fleet deployment, skipping BLE monitoring!"]];
+        return;
+    }
     // Note: We don't need to run `RequestAlwaysAuthorization`, already set in SensorControlForegroundDelegate.m.
     if ([CLLocationManager isMonitoringAvailableForClass:[CLBeaconRegion class]]) { // May be unecessary
         // Match all beacons with the specified UUID (This _must_ be same for all OpenPATH Deployments)
@@ -60,7 +65,8 @@ static NSString* const GEOFENCE_LOC_KEY = @"CURR_GEOFENCE_LOCATION";
         CLBeaconRegion *region = [[CLBeaconRegion alloc] initWithUUID:proximityUUID identifier:OpenPATHBeaconIdentifier];
         [locMgr startMonitoringForRegion:region];
         
-        NSLog(@"Started Monitoring BLE Region %@ during transition %@ ", region, transition);
+        [LocalNotificationManager addNotification:
+         [NSString stringWithFormat:@"Started Monitoring BLE Region %@ during transition %@ ", region, transition]];
         return;
     }
     // TODO: Add Error Handling:
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "cordova-plugin-em-datacollection",
-  "version": "1.9.5",
+  "version": "1.9.6",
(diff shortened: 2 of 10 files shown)


shankari:
@JGreenlee @Abby-Wheelis in case you want to review before I merge

src/ios/Location/TripDiaryActions.m
```diff
@@ -51,6 +51,11 @@ + (void) stopTracking:(NSString*) transition withLocationMgr:(CLLocationManager*
 }
 
 + (void)startBLEMonitoring:(NSString*) transition withLocationMgr:(CLLocationManager*)locMgr {
+    if (![ConfigManager isFleet]) {
+        [LocalNotificationManager addNotification:
+         [NSString stringWithFormat:@"Not a fleet deployment, skipping BLE monitoring!"]];
+        return;
+    }
```
JGreenlee, line 58: So this is the main change, yes? And everything else is just refactoring?

src/ios/Location/TripDiaryActions.m
```diff
@@ -60,7 +65,8 @@ + (void)startBLEMonitoring:(NSString*) transition withLocationMgr:(CLLocationMan
         CLBeaconRegion *region = [[CLBeaconRegion alloc] initWithUUID:proximityUUID identifier:OpenPATHBeaconIdentifier];
         [locMgr startMonitoringForRegion:region];
         
-        NSLog(@"Started Monitoring BLE Region %@ during transition %@ ", region, transition);
+        [LocalNotificationManager addNotification:
+         [NSString stringWithFormat:@"Started Monitoring BLE Region %@ during transition %@ ", region, transition]];
```
JGreenlee, line 69: This should help with any future debugging

JGreenlee (commented):
LGTM!
Has there been an opportunity to test this, or will we just do that on staging? We probably want to test both BLE and non-BLE cases so maybe we roll it out to dfc-fermata along with stage?

Abby-Wheelis:
>Has there been an opportunity to test this, or will we just do that on staging? We probably want to test both BLE and non-BLE cases so maybe we roll it out to dfc-fermata along with stage?

One of my test iPhones is configured with a `dfc-fermata` opcode. It hasn't tracked anything in a while, so I may need to check the battery status of the beacon in my car, but should be able to test the staging release with a BLE case.

src/ios/Location/TripDiaryActions.m
shankari: correct. I made sure to put this in a separate commit from the others

shankari:
Testing done:

- fleet mode
   - changed a local copy of dfc-fermata to connect to the e-mission-server on localhost
   - logged in using a newly made-up dfc-fermata opcode
   - tested the refactor
      - on iOS, used the debugger to verify that the `isFleet` call returned true
     - on android, I don't have the debugger set up with VSCode, but verified that when an `exited_geofence` was generated using the UI, it didn't change state, so it must have detected fleet more correctly
   - tested the new code
      - on iOS, added a breakpoint to the code that start scanning for BLE. It does not early return, but does not initiate the scan either since the emulator does not support beacons. But the new functionality (skip for non-fleet) has not caused a regression

- non-fleet mode
   - logged into to `dev-emulator-study` and restarted the app
   - tested the refactor
      - on iOS, used the debugger to verify that the `isFleet` call returned false
      - on android, verified that when an `exited_geofence` was generated, it changed state to `ongoing_trip`
   - tested the new code
      - on iOS, verified using the breakpoint that it early returns from `startBLEMonitoring`

Although... one issue that I appeared to run into was that the new config wasn't picked up on android until I reinstalled the app; will double check and file a new issue for that.

I am going to go ahead and merge this change and get it onto staging...

